PR #4, opened 2015-01-27
Introduce hawkular-parent as parent

ppalaga:


.travis.yml
```diff
@@ -3,7 +3,11 @@ language: java
 jdk:
   - oraclejdk8
 
-install: mvn clean install
+install:
+  - mvn -s .travis.maven.settings.xml -version -B
```
pilhuhn: Should we try to include -fae (fail at end), so that the CI build tries to do as much as possible, which could allow to uncover more than the first issue in a push?

ppalaga: I'd gladly say "no problem with -fae" if the whole PR was not outdated due to other PRs being merged faster. Nevertheless, I can say that sadly...

PR #8, opened 2015-01-29
HAWKULAR-12: adding tslint to the console build

vrockai:
I've added tslint as a part of our build. Right now it's a part of the "build" task and the "watch" task (with a separate configuration for watch, so the server doesn't stop on error).

I've used the "default" rules file from:
https://github.com/palantir/tslint/blob/master/tslint.json

With those rules setting, the build is failing. I didn't want to touch our ts code yet, so I've just prefixed those rules with underscores. We can later discuss what coding style we want to use and edit the rules afterwards.


mtho11:
@vrockai  could you also do this for hawkular-ui-components?

vrockai:
sure, will do

PR #9, opened 2015-01-30
Moving the node build to the target dir

vrockai:
So it doesn't mess with the maven stuff ppalaga did.

### DIFF
--- a/ui/console/pom.xml
+++ b/ui/console/pom.xml
@@ -20,6 +20,31 @@
 
     <build>
         <plugins>
+            <plugin>
+                <artifactId>maven-resources-plugin</artifactId>
+                <version>2.6</version>
+                <executions>
+                    <execution>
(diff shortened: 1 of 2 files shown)


ui/console/pom.xml
```diff
@@ -79,4 +104,39 @@
             </plugin>
         </plugins>
     </build>
+
+    <profiles>
+        <profile>
+            <id>dev</id>
```
pilhuhn: So only when I specify -Pdev, the stuff is not deleted? Would it make sense the other way around?

vrockai: dev profile is default. only if you specify -Prelease, the stuff is deleted. if you specify nothing, the downloaded node stuff stays in target directory.

mtho11:
Should probably add some instruction to READM.adoc

PR #10, opened 2015-02-02
HAWKULAR-20: Add navbar banner to the console

ammendonca:


mtho11:
@ammendonca can we add an image of some sort, just so we are able to add assets such as images, etc to the hawt.io 2 framework.

PR #13, opened 2015-02-03
building a prototype integration distribution

jmazzitelli:


pilhuhn:
Looks to me, like you have to manually build the `ui/console` first. And then the license checker is choking on all the (generated?) javascript files

[ERROR] Failed to execute goal com.mycila:license-maven-plugin:2.8:check (default) on project hawkular-console: Some files do not have the expected license header -> [Help 1]

integration-distro-test/pom.xml
```diff
@@ -0,0 +1,177 @@
...
+  <name>Hawkular Integration Distribution</name>
+
+  <properties>
+    <checkstyle.suppressions.file>${basedir}/src/main/scripts/checkstyle-suppressions.xml</checkstyle.suppressions.file>
+
+    <!-- The versions of the components that go into our integration distribution -->
+    <version.org.hawkular.nest>1.0.0.Alpha1-SNAPSHOT</version.org.hawkular.nest>
+    <version.org.hawkular.inventory>1.0-SNAPSHOT</version.org.hawkular.inventory>
+    <version.org.hawkular.alerts>1.0.0.Alpha1-SNAPSHOT</version.org.hawkular.alerts>
+    <version.org.rhq.metrics>0.2.8-SNAPSHOT</version.org.rhq.metrics>
```
pilhuhn: Looks like there is no 0.2.8-SNAPSHOT. Can we use the published 0.2.7 for the moment? Or fall-forward on 0.3.0-SNAPSHOT from Hawkular?

integration-distro-test/src/main/distro-resources/standalone/configuration/standalone.xml
```diff
@@ -0,0 +1,497 @@
...
+            </servlet-container>
+            <handlers>
+                <file name="welcome-content" path="${jboss.home.dir}/welcome-content"/>
+            </handlers>
+            <filters>
+                <response-header name="server-header" header-name="Server" header-value="WildFly/8"/>
+                <response-header name="x-powered-by-header" header-name="X-Powered-By" header-value="Undertow/1"/>
+            </filters>
+        </subsystem>
+        <subsystem xmlns="urn:jboss:domain:webservices:1.2">
```
pilhuhn: Should we remove that?

pilhuhn:
Standalone Cassandra seems to listen on port 7000 - can we try to change that port for the embedded one, so that one can have both?

pilhuhn:
Apparently there is still the dependency issue where metrics starts before C* is ready

integration-distro-test/pom.xml
jmazzitelli: We need 0.2.8-SNAPSHOT. 0.2.7 had a bug in metrics that I fixed and Stefan merged into metrics yesterday.

integration-distro-test/src/main/distro-resources/standalone/configuration/standalone.xml
jmazzitelli: We could probably remove ALOT and change ALOT in here. I left most of standalone.xml alone keeping all its default settings unless there was a need to change it to get something to work. Once we want to tighten up the server, there could be many things we will want to remove and change and tweak.

jmazzitelli:
If metrics starts before C* is ready, it means we need it to wait even longer. That's the bug I tried to fix in 0.2.8-SNAPSHOT. We could easily change wait loop so it waits longer. I guess you machine is even slower than mine :)

pilhuhn:
Ah, that waiting thingy is my bad then.

PR #18, opened 2015-02-13
Add a Pinger component and describe an end2end flow

pilhuhn:
This PR consists of 2 parts
- modules/end-to-end-test/  setup for (automated) testing of a flow of things through the Hawkular system

- Pinger : a component that does http HEAD requests to resources and sends the resulting "ping" times + result code to hawkular-metrics + a topic for alerts.
The pinger receives updates about resource add/remove from the inventory via a messaging topic.


pilhuhn:
Ok, Travis is now happy with dependencies available and so on.

modules/end-to-end-test/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+    <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>inventory-api</artifactId>
+      <version>1.0-SNAPSHOT</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.rhq.metrics</groupId>
+      <artifactId>rhq-metrics-api</artifactId>
+      <version>0.2.6</version>
```
jmazzitelli: we should use SNAPSHOT here for now right?

jmazzitelli:
I see the poms use many non-SNAPSHOT versions. Is the intention for this pinger to be just throw-away code for prototyping? If so, we can just commit this as-is without worrying about correctness here.

PR #31, opened 2015-02-26
sample module to build api jars for JSON encoding/decoding

jmazzitelli:
This can be used to start our "glue code" for JSON messaging. This shows an example of how to build an API jar that includes JSON schema file and the generated Java POJOs that correspond to that schema.

If you change the schema, and rebuild the mvn module, the generate POJOs are updated to match the schema.


theute:
Should we close and not merge ?

ppalaga:
Apparently this is/was not intended to be merged, thus closing.

PR #37, opened 2015-03-03
Added integration with hawkular accounts

jpkrohling:
This PR integrates Hawkular Accounts backend into Kettle. The change itself should not be invasive, except for the fact that a `standalone.conf` is to be shipped with it. It is required as we need to add a property to the command line specifying which realm Keycloak should import at boot.


jpkrohling:
@jmazzitelli , could you please review this one ?

PR #40, opened 2015-03-04
Changed assembly output: dir/zip/tar.gz by default, no dir for production

jpkrohling:


jpkrohling:
The build has failed due to a connection timeout for repository.jboss.org.  Would it be possible to re-trigger it? I can't see the "restart build" button for this repository, so, I guess I don't have permissions to do so.

theute:
Unfortunately, AFAIK Travis only allows us to restart the last ever build (from a PR, from a branch...)

PR #41, opened 2015-03-05
HAWKULAR-28: Initial prototype for Angular Sidebar menu directive

ammendonca:


mtho11:
Since we are nearing a release, I'm going to wait to merge this until after the release.

Also, I think the dual navigation could be confusing and needs further discussion with the UXD folks. We can have that talk on Monday's meeting.

mtho11:
Another option could be to remove this from Hawkular Console and place the directives in hawkular-ui-components while eliminating all second level navigation for the plugin.

mtho11:
sidebar-nav
Reference Image of sidebar nav.

mtho11:
Ideally, having two cooperating directives would provide a very flexible solution.

````html
<sidebarNav>
<navItem url="metrics/addUrl" label="Add Url" fa-font-image="" action="vm.addUrl()" />
<navItem url="metrics/addUrl" label="Add Url" fa-font-image="" action="vm.addUrl()" />

</sidebarNav>
````
I probably dont have all the correct properties or names, but just as example.

theute:
The home page when no URL has ever been entered should just have the URL input text (no navigation):
image

Then the left nav bar should appear and redirect to the "response" navigation entry with the content of http://localhost:8080/metrics/metricsResponse

It should not have top level bars. Alerts should be per URL (for now) and on the "alerts" side navigation bar. (Put all in there for now, dashboard, notifiers and definitions)  
We should IMO even hide the inventory tab (an idea if we need access to it temporarily for development/testing is to put it on a URL such as localhost:8080/dev non visible to the user).

image

Once this is done, we'll be closer from the UXD reqs and we can share for comments, better split of what is "done" vs what is "temporary".

PS:We are missing a screen to CRUD Urls, in the meantime we can put a link in the very top bar, close to the username to add a new URL and present the same screen as on first call when no URL is present.

ammendonca:
This PR is now obsolete. New PRs have been made (https://github.com/hawkular/hawkular-ui-components/pull/30 and https://github.com/hawkular/hawkular/pull/47).

PR #45, opened 2015-03-10
Added Hawkular Accounts UI integration

jpkrohling:


kettle/src/main/resources/standalone/configuration/hawkular-realm.json
```diff
@@ -57,17 +57,19 @@
...
-      "name": "hawkular-accounts-ui",
+      "name": "hawkular-ui",
       "enabled": true,
       "bearerOnly" : false,
       "publicClient": true,
-      "secret": "${uuid.hawkular.accounts.ui}",
+      "secret": "${uuid.hawkular.ui}",
       "fullScopeAllowed" : false,
       "redirectUris": [
+        "http://localhost:8080/*",
```
mtho11: Is this going to work inside containers?

jpkrohling: You mean, docker and related? Probably won't badly affect it, as this is a white-list of URLs that are accepted as "redirectUris". So, it's just a matter of adding new URLs to the list.

Jiri-Kremser: If the port is hard coded it doesn't work with docker when mapping the port 8080 to a different port. That's actually a suggested way to run it (see http://www.hawkular.org/docs/user/getting-started.html).

@jpkrohling can't be the port automagically inferred somehow? I remember I had similar issue with GWT debug mode and KeyCloak. (GWT debug mode was running on different port than production app)

jpkrohling: I'll check if there's a way to infer the port, but I'm afraid that would not be possible. What we can do is to add a set of ports that are "common" and add a documentation on how an administrator can add other addresses. 

In fact, having it "localhost" has already shown to be insufficient, as the continuous deployment server should be added to the list.

pilhuhn: Also that does not help. Just deploy that thing via `docker run -P ....  ` and you will get a "random" port and also a "random" IP depending on where you deploy it.

In JavaScript one can apparently use window.location() or such. And http headers seem to carry the "Origin" header (from Cors?) that gives the values.

jpkrohling: This JSON is to import the realm into Keycloak. This file is assembled at build time on Kettle and imported during Wildfly's first boot by Keycloak. The IP and Port mentioned on this list are the allowed host/port to which Keycloak will redirect back after login. So, if an application has requested the user to be redirected back to "acme.org/metrics", then "acme.org/*" needs to be on this list. Should an application ask the user to be redirected to "evil-acme.org/metrics", Keycloak will show an error message. This list can be changed at runtime directly at Keycloak, but we need a "starting point" (which is this realm.json file). 

In other words, this means:

* Docker internal IP/port is irrelevant, as what matters is what the browser will see.
* By default, we know only a few IPs/ports: localhost:8080 for Kettle, localhost:2772 for plugin development, localhost:9000 for a different javascript tool and possibly a couple of others. 
* For a production deployment, an admin will need to either change this realm.json or to use an external Keycloak server. In any case, this list needs to be fine-tuned to the environment, as localhost will never be a real production address for Hawkular.

pilhuhn: I wonder if this could be worked around by something like internally in Hawkular/Accounts always using a combo of some virtual host/port, which then is mapped by Docker itself via properties or such to the used ones. 

The issue is that e.g. for the CD we will never know in advance what ip/port we will get. Or when users just `docker run` hawkular. So we need the help of Docker to install the correct values.

jpkrohling: I played with it a bit, and it seems that our best option (at least for the moment) is to entirely remove the host/port parts, leaving only the path. This implies, however, that KC is running on the same host/port. As this will be the most common case, I think this is a good compromise.

Jiri-Kremser:
We can add ``--env HOST="`hostname -s`"`` when running the docker and leverage the `$HOST` variable inside the container. Not sure how though, perhaps docker could have a special start script that will do some sed `s;http://localhost;$HOST;` on the realm.json and then run the WF.

kettle/src/main/resources/standalone/configuration/hawkular-realm.json
Jiri-Kremser: We can add ``--env HOST="`hostname -s`"`` when running the docker and leverage the `$HOST` variable inside the container. Not sure how though, perhaps docker could have a special start script that will do some sed `s;http://localhost;$HOST;` on the realm.json and then run the WF.

Jiri-Kremser: Other workaround would be adding `--net="host"` to use the host network stack inside the container

jpkrohling: I believe the PR https://github.com/hawkular/hawkular/pull/46 address most of the common cases, including the scenario where Keycloak is on the same ip/port as Hawkular. Any other situation is, as I see it, a special (production?) deployment, and a fine tune on this will be required anyway. This fine tune can happen either directly on the embedded Keycloak via the admin UI, or by using an external Keycloak (which we'll have no control of).

PR #48, opened 2015-03-11
fix-url

pilhuhn:
Adapt the URLs in the scenario to real life


pilhuhn:
@jmazzitelli can you have a look please?

PR #51, opened 2015-03-17
HAWKULAR-49  - prevent timeout on add()

pilhuhn:
HAWKULAR-49  Do not start a ping on add, as this may block the add()
method for too long. Instead ping every 20 seconds.


mtho11:
The 20 seconds pings will demo much better than 1 minute pings.

PR #55, opened 2015-03-19
Get pinger back to work after metrics-api changes

pilhuhn:
Hawkular-Metrics has changed its api which makes the pinger fail.
This PR changes the data sent to follow the change in metrics.


mtho11:
Merged this in because I needed it in order to gather data so I can see it in charts (so I can fix the charts metrics api as well).

PR #68, opened 2015-04-02
Updating maven build profiles

vrockai:
Profiles add/changed:
* -Plink: Is using bower link functionality for linking the local hawkular-ui-components bower package.
* -Pcache: Was the `dev` profile before. It's used to persist (not download) node.js related stuff (npm/bower packages). The profile was renamed so it doesn't interfere with the `dev` profile anymore which has much broader functionality (creating user, etc) these days.


mtho11:
This is a great change!

PR #70, opened 2015-04-07
Support maven 3.3.1

metlos:
Frontend-maven-plugin doesn't work with Maven 3.3.1 before version 0.0.23.

### DIFF
--- a/ui/console/pom.xml
+++ b/ui/console/pom.xml
@@ -84,7 +84,7 @@
       <plugin>
         <groupId>com.github.eirslett</groupId>
         <artifactId>frontend-maven-plugin</artifactId>
-        <version>0.0.20</version>
+        <version>0.0.23</version>
 
         <configuration>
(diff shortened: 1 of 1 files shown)


mtho11:
Waiting on answer from Hawkular-dev ML on "What version of Maven do we support?"

mtho11:
I think we would only do this if maven 3.2.5 also works with it. Which it seems it does not.

metlos:
I've tried mvn 3.2.5 + frontend-maven-plugin 0.0.23 and the whole hawkular/hawkular project builds fine. I can't comment on whether the UI remains functional, because I have not tried it and am not sure about the level of brokenness/feature completion in the UI as it stands.

So in another words as far as I am concerned frontend-maven-plugin **does** work with both mvn 3.2.5 and mvn 3.3.1.

mtho11:
The code base has changed too much to allow a merge without significant effort and since this is a one line change -- I'm just going to commit it.
Commit ID: 3327ec1

PR #69, opened 2015-04-02
Making pinger and end-to-end test compatible with the new inventory.

Jiri-Kremser:
**Don't merge yet**
this can be merged after inventory implementation start sending notifications about created resources on hawkular bus (currently it's missing)

PR #72, opened 2015-04-08
[HKWINVENT-24] WIP - port pinger to new inventory bus

metlos:
This builds on top of #69 and re-adds the ability of the inventory to inform about changes inside over the bus.

There are still some problems though, so don't merge yet.


metlos:
This depends on PR https://github.com/hawkular/hawkular-inventory/pull/46 in inventory and on the ability of inventory to create entities and add properties to them at the same time (which is pending).

Once both of those features are in, this should theoretically work.

mtho11:
Can you start the pinger immediately, so we start collecting data sooner than collection period. We would like to have the UI show results faster than the current collection period.

metlos:
I'm glad I managed to hack around the pinger enough to make it receive the new inventory messages, but I see no deeper into it ;)

@pilhuhn would you be able to comment on @mtho11's request?

Jiri-Kremser:
Closing this pr on behalf of PR #75 (it's based on it)

PR #75, opened 2015-04-13
Hwkinvent-24 - port pinger to new inventory bus + setup work

Jiri-Kremser:
this is pr #72 + the code responsible for creating the tenant, environment and resourceType. It conceptually fits the best to the Pinger.

This can be merged after the PRs in hawkular/ui-{services|components} are merged

### DIFF
--- a/modules/pinger/pom.xml
+++ b/modules/pinger/pom.xml
@@ -47,7 +47,7 @@
     <dependency>
       <groupId>org.hawkular.inventory</groupId>
       <artifactId>inventory-api</artifactId>
-      <version>1.0.0-SNAPSHOT</version>
+      <version>0.0.1-SNAPSHOT</version>
     </dependency>
 
     <dependency>
--- a/modules/pinger/pom.xml
+++ b/modules/pinger/pom.xml
@@ -52,7 +52,7 @@
 
     <dependency>
       <groupId>org.hawkular.inventory</groupId>
-      <artifactId>bus-api</artifactId>
+      <artifactId>inventory-bus-api</artifactId>
       <version>0.0.1-SNAPSHOT</version>
     </dependency>
 
--- a/modules/pinger/pom.xml
+++ b/modules/pinger/pom.xml
@@ -51,6 +51,12 @@
     </dependency>
 
     <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>bus-api</artifactId>
+      <version>0.0.1-SNAPSHOT</version>
+    </dependency>
+
+    <dependency>
       <groupId>org.hawkular.metrics</groupId>
       <artifactId>clients-common</artifactId>
       <version>0.3.0-SNAPSHOT</version>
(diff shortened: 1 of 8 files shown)


modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
```diff
@@ -65,31 +72,64 @@
...
-                    String type = (String) m.get("type");
-                    Map<String, String> params = (Map<String, String>) m.get("parameters");
-                    String url = params.get("url");
-                    destinations.add(new PingDestination(id, url));
+        int attempts = 0;
+        while (attempts++ < 10) {
+            try {
+                Client client = ClientBuilder.newClient();
+                // TODO: inventory does not have to be co-located
+                final String host = "http://localhost:8080";
```
pilhuhn: Pull out of loop? (+next line)

modules/pinger/src/main/java/org/hawkular/component/pinger/Log.java
```diff
@@ -27,14 +28,14 @@
  * @author Heiko W. Rupp
  */
 @MessageLogger(projectCode = "HAWKULAR")
-public interface Log {
+public interface Log extends BasicLogger {
```
pilhuhn: Just curious: why is that needed?

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
```diff
@@ -53,7 +59,8 @@
     // How long do we wait between each round
     private static final int WAIT_MILLIS = 500;
 
-    String tenantId = "test";
+    private final String tenantId = "test";
```
pilhuhn: Shouldn't those be retrieved from Inventory now that Inventory supports them?

Jiri-Kremser: someone has the create the default tenant and environment. This should be probably done by the keycloak ui code, this is just to make it working w/ the 'end-to-end' test

modules/pinger/src/main/java/org/hawkular/component/pinger/Log.java
Jiri-Kremser: It isn't (http://docs.jboss.org/seam/3/3.1.0.Final/reference/en-US/html/solder-logging.html#jbosslogging), but if you do that, you'll get the generic methods like debug, trace, logf, etc.

If we want to be super strict and allow only the predefined (and localized) methods than it can go away. Btw. alert does it the same way.

modules/pinger/pom.xml
```diff
@@ -47,7 +47,13 @@
     <dependency>
       <groupId>org.hawkular.inventory</groupId>
       <artifactId>inventory-api</artifactId>
-      <version>1.0.0-SNAPSHOT</version>
+      <version>0.0.1-SNAPSHOT</version>
+    </dependency>
+
+    <dependency>
+      <groupId>org.hawkular.inventory</groupId>
+      <artifactId>bus-api</artifactId>
```
metlos: This needs to be changed to "inventory-bus-api" now that https://github.com/hawkular/hawkular-inventory/pull/49 is merged.

Jiri-Kremser: done

PR #80, opened 2015-04-15
HAWKULAR-64 - Added logout indicator

jpkrohling:


mtho11:
Merging even though there were errors, because it needed a bump from hawkular-ui-components.

PR #81, opened 2015-04-16
Console: Fix user dropdown needing to be clicked twice

ammendonca:


ui/console/src/main/webapp/index.html
```diff
@@ -47,7 +47,7 @@
         <div class="collapse navbar-collapse navbar-collapse-1">
           <ul class="nav navbar-nav navbar-utility pull-right">
             <li class="dropdown">
-              <a href="#" class="dropdown-toggle" data-toggle="dropdown">
+              <a href="#" dropdown-toggle>
```
mtho11: Is 'dropdown-toggle' an Angular directive?

PR #92, opened 2015-05-07
HAWKULAR-135 Pinger is completely trusting each SSL certificate

marvec:


marvec:
Ok, hawkular pinger was successfully built. The other module failed.

ppalaga:
Welcome, Man in the Middle!

ppalaga:
I checked locally that the tests are passing. The code is OK for the present state of the project, but we should not trust every cert in the future. I am OK with merging.

ppalaga:
I added a follow-up task: https://issues.jboss.org/browse/HAWKULAR-203

PR #97, opened 2015-05-13
HAWKULAR-144 Make pinger support GET requests

ppalaga:


modules/pinger/src/test/java/org/hawkular/component/pinger/test/PingerTest.java
```diff
@@ -34,27 +35,51 @@
...
+        Assert.assertEquals(200, status.getCode());
+        Assert.assertFalse(status.isTimedOut());
+
+    }
+
+    @Test
+    public void testHeadPinger() throws Exception {
+
+        Pinger pinger = new Pinger();
+        PingDestination destination = new PingDestination("123","http://hawkular.github.io", "HEAD");
```
pilhuhn: Should we introduce / use constants instead of `"HEAD"` and so on?

pilhuhn:
I think we also need a followup Jira to bring this change to the UI

modules/pinger/src/test/java/org/hawkular/component/pinger/test/PingerTest.java
ppalaga: I do not see such a need in this test. 
In `PingDestination`, there is a constant for the default method `"GET"` and there is no other method named in non-test code.

ppalaga:
+1 for a followup Jira to bring this change to the UI

PR #100, opened 2015-05-18
Hwkinvent 38+hawkular 100

Jiri-Kremser:
3rd attempt :) Because all the previous PRs, even though they were merged, they didn't manage to get into Hawkular upstream, because of the repositories reorganization.


mtho11:
LGTM. I would like to see more type safety around resource type id, tenant id, and metric id  via: https://github.com/hawkular/hawkular/blob/fa8e2a4dd60e9d2373b5eb9fe58e148c1dbcaaeb/ui/console/src/main/scripts/plugins/metrics/plugins/metrics/ts/metricsTypes.ts

PR #102, opened 2015-05-19
HAWKULAR-200: Rework/simplify the build

theute:
This is all about HAWKULAR-200.
Kettle is dead, welcome "dist"... Not as exciting but more familiar.
This no longer depends on nest-distro (for good and bad).

It will likely conflict with https://github.com/hawkular/hawkular/pull/101 I can rebase once 101 is merged if needed


theute:
Note: I didn't test the WF agent, I tested docker before the rebase (didn't do after)

I still need to update the doc + website and someone will likely have to update the scripts for the live instances.

dist/assembly.xml
```diff
@@ -19,57 +19,75 @@
 -->
 
 <assembly>
-  <id>distribution</id>
+  <id>zip-distrib</id>
+  <formats>
```
jmazzitelli: I don't think <formats> here is needed. I believe Peter's last PR removed this but somehow the rebase for this PR brought it back. The pom.xml's assembly plugin configuration defines zip and tar.gz already.

dist/assembly.xml
```diff
@@ -19,57 +19,75 @@
...
+      <directoryMode>0755</directoryMode>
+    </fileSet>
+    <fileSet>
+      <directory>src/main/resources/wildfly/bin</directory>
+      <outputDirectory>/${wildfly.dist.zip.root.dir}/bin</outputDirectory>
+      <filtered>false</filtered>
+      <fileMode>0744</fileMode>
+      <directoryMode>0755</directoryMode>
+    </fileSet>
+     <!-- non-binary files -->
```
jmazzitelli: indentation is incorrect. Not sure why checkstyle didn't catch it.

PR #103, opened 2015-05-20
Remove duplicate properties

ppalaga:


ppalaga:
@theute could you please restart travis once again?

PR #105, opened 2015-05-20
Upgrade to hawkular-parent 13

ppalaga:


theute:
I did it as part of https://github.com/hawkular/hawkular/pull/104 (to fix the issue with assembly plugin). So let's close this one and apply the other one if you don't mind.

PR #106, opened 2015-05-20
[HWKINVENT-36] - Modifications to dist to support accounts-integrated inventory.

metlos:
This should be merged only once https://github.com/hawkular/hawkular-inventory/pull/63 is merged.

### DIFF
--- a/dist/assembly.xml
+++ b/dist/assembly.xml
@@ -90,7 +90,7 @@
       <outputDirectory>${wildfly.dist.zip.root.dir}/modules/org/hawkular/nest/main/deployments</outputDirectory>
       <useProjectArtifact>false</useProjectArtifact>
       <includes>
-        <include>org.hawkular.inventory:inventory-rest-api</include>
+        <include>org.hawkular.inventory:hawkular-integrated-inventory-rest</include>
         <include>org.hawkular:hawkular-pinger</include>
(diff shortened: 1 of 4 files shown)


metlos:
This can now be merged with https://github.com/hawkular/hawkular-inventory/pull/63 in inventory master.

@mtho11: please note that doing that will probably break UI.

There no longer is `/tenants` endpoint, because it is no longer possible to list all tenants. Instead a new `/tenant` endpoint is added that can be used to get/modify info about the current tenant which has the same ID as the current persona.

All the UI code calling into Inventory's REST should therefore replace the "test" tenant with the ID of the current persona.

The setup of the "test" environment and the "URL" resource type and appropriate metric types is automagic and happens behind the scenes.

mtho11:
Its failed travis: because of: [ERROR] Failed to execute goal on project hawkular-dist: Could not resolve dependencies for project org.hawkular:hawkular-dist:pom:1.0.0-SNAPSHOT: Could not find artifact org.hawkular.inventory:hawkular-integrated-inventory-rest:war:0.0.2-SNAPSHOT in jboss-public-repository-group (https://repository.jboss.org/nexus/content/groups/public-jboss/)

Someone with privs need to kick Travis off again to see if the inventory snapshot eventually shows up.

theute:
Merged

PR #109, opened 2015-05-20
[HAWKULAR-215] - Use shared C* instance using wfly cassandra subsystem

metlos:
This requires https://github.com/hawkular/wildfly-cassandra/pull/5.


pilhuhn:
What is the current state here?

metlos:
I think this can be closed as the JIRA (https://issues.jboss.org/browse/HAWKULAR-215) has been rejected in favor of using the cassandra EAR that we already have. We may revisit that decision later.

ppalaga:
Closing. Cassandra EAR from commons is in use now.

PR #111, opened 2015-05-21
HAWKULAR-228 - Updated Keycloak to 1.2.0.Final

jpkrohling:


jpkrohling:
When trying to add an URL, I get the following message: "Error during saving metrics. [object Object]". By looking at the developer console from the browser, I don't see any indications that this is caused by this update. But before merging, would be nice to confirm that this is a known issue.

PR #112, opened 2015-05-21
HAWKULAR-234 - Renamed Hawkular Persona HTTP header

jpkrohling:


jpkrohling:
As this is a breaking change, I bumped accounts before merging a similar change on that side, so, upon merging of this, there *should not* be any incompatibilities.

PR #113, opened 2015-05-21
HAWKULAR-173 Collect traits from a URL

ppalaga:


modules/pinger/pom.xml
```diff
@@ -35,9 +35,19 @@
   <description>Component that allows to ping remote web sites and report their status code + response time
   </description>
 
+  <properties>
+    <version.org.hawkular.inventory>0.0.1.HWKINVENT-36</version.org.hawkular.inventory>
```
metlos: ?

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
```diff
@@ -47,72 +50,114 @@
...
-     * @param metrics List of metrics
+     *
+     * @param tenantId
+     *            Name of the tenant
+     * @param metrics
+     *            List of metrics
      */
     @Asynchronous
-    public void sendToMetricsViaRest(String tenantId, List<Map<String, Object>> metrics) {
+    private void sendToMetricsViaRest(String tenantId, List<Map<String, Object>> metrics) {
```
metlos: I'm not sure about the effect of `@Asynchronous` on a private method. Does it still work?

modules/pinger/pom.xml
ppalaga: You see? You are even faster than Travis :)

modules/pinger/src/main/java/org/hawkular/component/pinger/PingDestination.java
```diff
@@ -16,74 +16,165 @@
...
-    public PingDestination(String resourceId, String url, String method) {
+    public PingDestination(String tenantId, String environmentId, String resourceId, String url, String method) {
+        this.tenantId = tenantId;
+        this.environmentId = environmentId;
         this.resourceId = resourceId;
         this.url = url;
         this.method = method == null ? DEFAULT_METHOD : method;
     }
 
+    /** Compares only {@link #resourceId} for performance reasons.
```
metlos: This is technically incorrect, because resourceId is guaranteed to be unique only within its environment. I.e. in current situation, two tenants could create a resource with the same id (not sure if that's possible) and they would be claimed equal.

modules/pinger/pom.xml
ppalaga: Force-pushed a fix.

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
```diff
@@ -16,114 +16,189 @@
...
+    @Observable
+    private Inventory.Mixin.Observable inventory;
+
+    final NewUrlsCollector newUrlsCollector = new NewUrlsCollector();
 
-    private final PingerConfiguration configuration = PingerConfiguration.defaults();
+    private void addUrl(String tenantId, String envId, String url) {
+        Log.LOG.infof("Putting url to inventory: %s", url);
+
+        Single tenant = null;
```
metlos: This whole setup block for tenant, env, and resource type should be unnecessary, because these are set up automagically by glue (hawkular-integrated-inventory-rest/TemporaryHacks).

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
ppalaga: You seem to be right http://stackoverflow.com/a/3815844

modules/pinger/src/main/java/org/hawkular/component/pinger/PingDestination.java
ppalaga: Good to know, thanks!

modules/pinger/src/main/java/org/hawkular/component/pinger/TraitsPublisher.java
```diff
@@ -0,0 +1,62 @@
...
+    public void publish(PingStatus status) {
+        final Traits traits = status.getTraits();
+        Builder resourceBuilder = Resource.Update.builder();
+        resourceBuilder.withProperty("traits-collected-on", traits.getTimestamp());
+        for (Entry<TraitHeader, String> entry : traits.getItems().entrySet()) {
+            resourceBuilder.withProperty("trait-" + entry.getKey().toString(), entry.getValue());
+        }
+
+        PingDestination dest = status.getDestination();
+        inventory.tenants().get(dest.getTenantId()).environments().get(dest.getEnvironmentId()).feedlessResources()
```
metlos: This is not correct.

`Entity.update()` is meant to create a new instance of an entity with the internals updated from the provided update object. But that's just the in-memory instance, it is not connected or persisted to the backend store in any way.

Use should use this instead:
```java

inventory.tenants().get(dest.getTenantId()).environments().get(dest.getEnvironmentId()).feedlessResources().update(dest.getResourceId(), resourceBuilder.build());
```

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
ppalaga: Nevermind, the code is thought to be there only as long as the UI for adding urls is broken.

modules/pinger/src/main/java/org/hawkular/component/pinger/TraitsPublisher.java
ppalaga: Thanks, fixed

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
```diff
@@ -47,72 +50,110 @@
...
 
-    @javax.annotation.Resource( lookup = "java:/topic/HawkularMetricData")
+    private static void addDataItem(List<Map<String, Object>> mMetrics, String resourceId, long timestamp,
+            Number value, String name) {
+        Map<String, Number> dataMap = new HashMap<>(4);
+        dataMap.put("timestamp", timestamp);
+        dataMap.put("value", value);
+        List<Map<String, Number>> data = new ArrayList<>(1);
+        data.add(dataMap);
+        Map<String, Object> outer = new HashMap<>(4);
```
jpkrohling: Why 4, if you only add 2?

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
```diff
@@ -47,72 +50,110 @@
...
-                data.put("data",metrics);
+                Map<String, Object> outer = new HashMap<>(1);
 
-                outer.put("metricData",data);
+                Map<String, Object> data = new HashMap<>(2);
+                data.put("tenantId", status.getDestination().getTenantId());
+                data.put("data", singleMetrics);
 
-                Endpoint endpoint = new Endpoint(Endpoint.Type.TOPIC,topic.getTopicName());
+                outer.put("metricData", data);
```
jpkrohling: This seems very similar to the addDataItem. Perhaps a comment here (or there, or somewhere) on how the data is structured?

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
```diff
@@ -16,114 +16,189 @@
...
+        }
+
+        /**
+         * Returns the list of {@link PingDestination}s collected by this {@link NewUrlsCollector}. It is safe to call
+         * this method concurrently from any random thread.
+         *
+         * @return the list of {@link PingDestination}s
+         */
+        public List<PingDestination> getNewUrls() {
+            synchronized (this) {
```
jpkrohling: You could synchronize the whole method, no?

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
```diff
@@ -16,114 +16,189 @@
...
-                Thread.currentThread().interrupt();
-                return;
-            }
+        /*
+         * Add the observer before reading the existing URLs from the inventory so that we do not loose the URLs that
+         * could have been added between those two calls.
+         */
+        inventory.observable(Interest.in(Resource.class).being(Action.created())).subscribe(newUrlsCollector);
+
+        Set<Resource> urls = inventory.tenants().getAll().resourceTypes().getAll(With.id(PingDestination.URL_TYPE))
```
jpkrohling: Does this results in a REST call? If so, it might fail, as it might run before inventory (or keycloak) has finished starting.

modules/pinger/src/main/java/org/hawkular/component/pinger/Traits.java
```diff
@@ -0,0 +1,190 @@
...
+ */
+public class Traits {
+
+    /**
+     * The enumeration of HTTP response headers that are considered to bear interesting information about a monitored
+     * site.
+     */
+    public enum TraitHeader {
+        SERVER("server"), X_ASPNET_VERSION("x-aspnet-version"), X_POWERED_BY("x-powered-by"), X_RUNTIME("x-runtime"),
+        X_VERSION("x-version");
```
jpkrohling: Is there a list somewhere of such interesting headers? I'd include some from CloudFlare there (CF-RAY), for instance. This is helpful for people who have content behind CloudFlare to identify where the problem might be. There are also perhaps some OpenShift headers that could be useful?

modules/pinger/src/main/java/org/hawkular/component/pinger/Traits.java
```diff
@@ -0,0 +1,190 @@
...
+     * @param timestamp the UNIX timestamp when the response was received
+     * @return a new {@link Traits}
+     */
+    public static Traits collect(HttpResponse httpResponse, long timestamp) {
+        Builder<TraitHeader, String> builder = new ImmutableMap.Builder<TraitHeader, String>();
+
+        HeaderIterator headers = httpResponse.headerIterator();
+        while (headers.hasNext()) {
+            Header header = headers.nextHeader();
+            TraitHeader traitHeader = TraitHeader.fastValueOf(header.getName());
```
jpkrohling: It would be interesting to test this with Java's stream API. Cleaner code, and *probably* better performance.

modules/pinger/src/test/java/org/hawkular/component/pinger/PingerTest.java
```diff
@@ -75,39 +75,11 @@ public void testPostPinger() throws Exception {
     public void testSslPinger() throws Exception {
 
         Pinger pinger = new Pinger();
-        PingDestination destination = new PingDestination("123","https://www.perfcake.org");
+        PingDestination destination = newDestination("https://www.perfcake.org", "GET");
```
jpkrohling: Can you add a server making use of SNI as well? Example of a server that makes use of SNI: https://mail.kroehling.de . But don't add my address to the test, as this machine is in an unstable environment ,so, your test might fail without reason :)

jpkrohling:
I finished the review. I cannot review the business behind it, but the code looks OK to me. Some minor adjustments could be thought of, but nothing major.

modules/pinger/src/main/java/org/hawkular/component/pinger/Traits.java
pilhuhn: There is a list of headers we want to look for in the jira - or its "parent" jira.

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
metlos: Nope, this is a straight API call with no remote communication.

modules/pinger/src/main/java/org/hawkular/component/pinger/Traits.java
theute: We can add more later, what is interesting in our context for now, is to know:
X-Powered-By: Undertow/1
Server: WildFly/8

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
pilhuhn: How does that work then? Does it basically pull in a version of inventory and uses it in parallel to the one running as a service inside of Hawkular?

metlos: Yes, pinger is now required to run co-located with Hawkular. Doing it "properly" and enable remote pinger would be too much work for the next iteration. This will have to be revisited of course, but I think this is actually kinda good for the SaaS usecase, because using this approach we can circumvent the authz that is now in place on the REST endpoint. Being able to authz the pinger requires much more work on the pinger side as well as inventory side to do it "properly" (it could be made working with inventory as it is now, but in a little bit hacky way).

modules/pinger/src/main/java/org/hawkular/component/pinger/Traits.java
metlos: Not sure the overhead of running this through the streaming API is worth it given there is no more than a handful of headers in a request.

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
pilhuhn: The question was less about co-located with Hawkular (or inside the Hawkular-server), but rather about "when inventory is in one deployment and the pinger in a different one, do they "see each other"? Or is inventory now deployed inside pinger's lib/ directory as a new dependency?

metlos: Ok, I took Peter's word for it working, but this is a problem apparently.

Because pinger has the inventory as a dep and rest api, too, there are 2 singleton beans of inventory being initialized. This breaks with the in-memory tinkerpop impl we're using, because it tries to initialize from the same directory twice...

metlos:
There is a problem with the pinger and rest API each having its own copy of inventory database being persisted/read from the same directory.

modules/pinger/src/main/java/org/hawkular/component/pinger/Traits.java
jpkrohling: Yet another reason to use the standard Java API instead of an external lib? :-)

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
ppalaga: Yes, you are right. Having default fill factor 0.75, 4 was the smallest initial capacity for me that could hold 2 elements without rehashing. Firstly, 3 would actually suffice too, as {{3 * 0.75 = 2.25}} (which I was not able to eval in my head before) and secondly, HashMap in Java 8 seems to work differently than before, meaning that initial capacity of 2 would suffice on this place too.

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
ppalaga: Yes, the code here is not very clean. I have no idea if it works at all. I have just moved it here from other place. Putting the data together like this is not safe. It would be ideal if Metrics API offered some type-safe way to do this, like Inventory does. Maybe it does, I have not investigated. Anyway changing anything on writing to Metrics is not the objective of this PR.

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
ppalaga: Yes, changed to a dedicated lock field.

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
ppalaga: @pilhuhn yes, pinger deployment was made dependent on inventory deployment here: https://github.com/hawkular/hawkular/pull/113/files#diff-0ca99e5a5fa37d670c536a677376c0c7R24

modules/pinger/src/main/java/org/hawkular/component/pinger/Traits.java
ppalaga: -1, not worth experimenting at this stage

modules/pinger/src/test/java/org/hawkular/component/pinger/PingerTest.java
ppalaga: -1, out of the present scope. These tests pinging 3rd party machines in the wild is somewhat fragile, IMO. I'd vote for mocking the responses rather than adding further servers in foreign networks.

ppalaga:
Closing this one as it was already merged within PR#121 the followup stuff is in https://github.com/hawkular/hawkular/pull/125

PR #115, opened 2015-05-21
HAWKULAR-140 - Consuming the provider from ui-services for accounts.

jpkrohling:


ui/console/src/main/scripts/plugins/accounts/plugins/accounts/ts/organizations.ts
```diff
@@ -19,8 +19,8 @@
 module HawkularAccounts {
 
     export var OrganizationsController = _module.controller("HawkularAccounts.OrganizationsController", [
-        '$rootScope', '$scope', 'HawkularAccounts.OrganizationService', '$log', '$location',
-        ($rootScope, $scope, OrganizationService, $log, $location) => {
+        '$rootScope', '$scope', '$log', '$location', 'HawkularAccount',
+        ($rootScope, $scope, $log, $location, HawkularAccount) => {
 
             $scope.organizations = [];
```
mtho11: Angular best practices is not to use $scope at all. But instead, use 'this' and 'ControllerAs' syntax.
For reference: http://toddmotto.com/digging-into-angulars-controller-as-syntax/

jpkrohling: Created https://issues.jboss.org/browse/HAWKULAR-240 for this, as this would require more changes than this isolated part.

PR #119, opened 2015-05-22
Bumped inventory version

jpkrohling:


jpkrohling:
It seems @metlos was quicker than me :) See #118

PR #120, opened 2015-05-22
move away from snapshots for bus/nest/agent

jmazzitelli:


theute:
:+1: 
From a process perspective, it would be better to create a branch in your own clone (ie: not in https://github.com/hawkular )

PR #124, opened 2015-05-23
Pinger fixes:

metlos:
* use JNDI instead of CDI for inventory injection to prevent initialization
errors (kudos to jpkroehling for suggesting that)
* fix resource update when publishing traits

This depends on https://github.com/hawkular/hawkular-inventory/pull/69.

With both the inventory fixes and this PR in, it is possible to add URLs and monitor them successfully. Delete is still broken and will have to be fixed on the inventory side in a separate PR (there's non-trivial amount of work needed to make it work correctly).

### DIFF
--- a/modules/pinger/pom.xml
+++ b/modules/pinger/pom.xml
@@ -62,12 +62,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.hawkular.inventory</groupId>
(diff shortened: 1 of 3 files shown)


metlos:
Emergency merge to get this into repos so that others can start working with pinger on Monday.

PR #128, opened 2015-05-26
Add template realm as optional

lucasponce:
Adding hawkular-realm-for-dev.json in the hawkular-dist as optional file makes easy to populate a demo user when using the hawkular-dist from nexus.
If not, a pre-process to get the UUID is needed and makes things more difficult to start using hawkular from a pom.xml


jpkrohling:
:+1: , tested both the dev build and the regular build, looks fine on both, ie, the file is available on standalone/configuration for both variants.

PR #131, opened 2015-05-26
HAWKULAR-250: switched to xml2js lib for xml parsing

vrockai:


jshaughn:
This seems to solve the build issues on systems not able to build the native lib. Thanks!

PR #132, opened 2015-05-26
HAWKULAR-256 Move last updated timestamp to header 

mtho11:
instead of being repeated in table body.


mtho11:
Built fine. Travis stalled out. Merging.

PR #133, opened 2015-05-26
HK 238 Fix hawkular-end-to-end-tests

ppalaga:


pom.xml
```diff
@@ -24,7 +24,7 @@
   <parent>
     <groupId>org.hawkular</groupId>
     <artifactId>hawkular-parent</artifactId>
-    <version>13</version>
+    <version>14-SNAPSHOT</version>
```
metlos: is this necessary?

modules/end-to-end-test/src/test/groovy/org/hawkular/integration/test/Scenario1ITest.groovy
```diff
@@ -0,0 +1,143 @@
...
+        //def response = client.get(path: "/hawkular-accounts/organizations")
+        def response = client.get(path: "/hawkular-accounts/personas/current")
+        assertResponseOk(response.status)
+        String tenantId = response.data.id
+        // println "tenantId = $tenantId"
+
+        /* assert the test environment exists */
+        response = client.get(path: "/hawkular/inventory/$tenantId/environments/$environmentId")
+        assertResponseOk(response.status)
+        org.junit.Assert.assertEquals(environmentId, response.data.id)
```
metlos: why not static-import these, or at least not use the full class name?

metlos:
except for the 14-SNAPSHOT parent, this looks good to me. Let's see what Travis says ;)

modules/end-to-end-test/src/test/groovy/org/hawkular/integration/test/Scenario1ITest.groovy
ppalaga: Ehm... because editing Groovy kinda sucks. But yes, for the sake of the pure beauty, I can import them :)

pom.xml
ppalaga: Haha, the reason why Travis must fail

ppalaga:
@metlos, meanwhile I released parent 14. Could you please merge after Travis passes?

PR #135, opened 2015-05-27
HAWKULAR-257 Run end to end tests by Travis

ppalaga:


ppalaga:
@pilhuhn, could you please review?

Esp., I left the `itest` profile off by default, but I activated it for Travis. This is because the end to end tests start a vanilla WildFly instance on port 8080 that people often have in use by their own Hawkular instance.

`itest` can be made active by default later, when Hawkular is able to start with `-Djboss.socket.binding.port-offset` which is unfortunately not the case ATM.

PR #140, opened 2015-05-28
HAWKULAR-269 - Using jboss.bind.address for KC server URL

jpkrohling:
@theute  , could you please review this one? If approved, I'll also merge the following, as the changes are identical: https://github.com/hawkular/hawkular-accounts/pull/27

### DIFF
--- a/dist/src/main/resources/wildfly/patches/standalone.xsl
+++ b/dist/src/main/resources/wildfly/patches/standalone.xsl
@@ -759,7 +759,7 @@
         </auth-server>
         <realm name="hawkular">
           <auth-server-url>/auth</auth-server-url>
-          <auth-server-url-for-backend-requests>http://localhost:8080/auth</auth-server-url-for-backend-requests>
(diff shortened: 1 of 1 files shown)


theute:
I am not sure to be the correct reviewer, I was hoping that it would fix the issue when binding to an IP and I get the spinners forever for availability and response time but that makes no difference.

Basically before and after this change I see no difference in what I tried

jpkrohling:
The PR should fix only the accounts part of the problem. So, in the logs you should not see anymore a "failed to retrieve public key" or related issues if you start the server with "-b specific_ip". Although you might still see it if the server is starting up or shutting down.

pilhuhn:
@theute What spinners do you actually mean here?

theute:
@pilhuhn I think it's unrelated. When I bind to some IP, the availability and response time would keep display the waiting image without a value forever, I think this was fixed in some other commit but I didn't check

pilhuhn:
Ok, the spinners are not (no longer?) an issue with recent master and after HAWKULAR-106 ( #158 ) was merged into it.

ppalaga:
This fix is still needed, even after #158 . @pilhuhn could you please merge it?

PR #145, opened 2015-05-28
HAWKULAR-235 Make Pinger listen to URLs removed

ppalaga:


ppalaga:
@pilhuhn could you please review, once the CI passes?

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
```diff
@@ -119,20 +203,23 @@ public void call(Resource r) {
     @javax.annotation.Resource(lookup = "java:global/Hawkular/ObservableInventory")
     private Inventory.Mixin.Observable inventory;
 
-    final NewUrlsCollector newUrlsCollector = new NewUrlsCollector();
+    final UrlChangesCollector newUrlsCollector = new UrlChangesCollector();
```
pilhuhn: variable naming?

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
```diff
@@ -52,47 +55,128 @@
...
         }
+    }
 
+    /**
+     * Collects URL additions and removals reported by {@link PingManager#inventory} and synchronizes the various
+     * threads reporting the new URLs and those ones consuming them.
+     *
+     * @author <a href="https://github.com/ppalaga">Peter Palaga</a>
+     */
+    static class UrlChangesCollector {
```
pilhuhn: Should that go into its own class file to not "pollute" this manager class? Probably along with the `UrlChange`

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
ppalaga: +1

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
ppalaga: +1, even unit testing will be cleaner

modules/pinger/src/main/java/org/hawkular/component/pinger/UrlChangesCollector.java
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Collects URL additions and removals reported by {@link PingManager#inventory} and synchronizes the various threads
+ * reporting the new URLs and those ones consuming them.
+ *
+ * @author <a href="https://github.com/ppalaga">Peter Palaga</a>
+ */
+public class UrlChangesCollector {
+    /**
+     * An an add or remove operation that can be performed on a {@link Set}.
```
pilhuhn: Typo "An an"

modules/pinger/src/main/java/org/hawkular/component/pinger/UrlChangesCollector.java
```diff
@@ -0,0 +1,157 @@
...
+                    changes.add(new UrlChange(Action.created(), PingDestination.from(r)));
+                }
+            }
+        }
+    };
+
+    /**
+     * A callback for the {@link Inventory} that collects URL removals.
+     */
+    private final Action1<Resource> urlDeletedAction = new Action1<Resource>() {
```
pilhuhn: Is it possible to only use one action callback for both cases and then internally inside `call` differentiate on the incoming event?

ppalaga: No, `call` get only a `Resource` and one cannot tell whether it was added or removed.

PR #146, opened 2015-05-28
HAWKULAR-276 - only put jdoe in config for dev builds.

jmazzitelli:
only put jdoe in config for dev builds. production builds, the agent will be disabled by default and the user is required to enable it and configure it with their own user/pass credentials

### DIFF
--- a/dist/src/main/resources/wildfly/patches/standalone.xsl
+++ b/dist/src/main/resources/wildfly/patches/standalone.xsl
@@ -254,20 +254,31 @@
 
       <!-- Hawkular Monitor agent subsystem -->
       <subsystem xmlns="urn:org.hawkular.agent.monitor:monitor:1.0"
-                 enabled="true"
                  apiJndiName="java:global/hawkular/agent/monitor/api"
                  numMetricSchedulerThreads="3"
                  numAvailSchedulerThreads="3">
+        <xsl:if test="$kettle.build.type='dev'">
+          <xsl:attribute name="enabled">true</xsl:attribute>
+        </xsl:if>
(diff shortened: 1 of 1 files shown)


dist/src/main/resources/wildfly/patches/standalone.xsl
```diff
@@ -254,20 +254,31 @@
 
       <!-- Hawkular Monitor agent subsystem -->
       <subsystem xmlns="urn:org.hawkular.agent.monitor:monitor:1.0"
-                 enabled="true"
                  apiJndiName="java:global/hawkular/agent/monitor/api"
                  numMetricSchedulerThreads="3"
                  numAvailSchedulerThreads="3">
+        <xsl:if test="$kettle.build.type='dev'">
```
pilhuhn: Would it make sense to use xsl:choose instead? See https://msdn.microsoft.com/en-us/library/ms256169(v=vs.110).aspx

dist/src/main/resources/wildfly/patches/standalone.xsl
```diff
@@ -254,20 +254,31 @@
...
         <diagnostics enabled="true"
                      reportTo="LOG"
                      interval="1"
                      timeUnits="minutes"/>
 
-        <storage-adapter type="HAWKULAR"
-                         username="jdoe"
-                         password="password">
-                         <xsl:attribute name="url">http://&#36;{jboss.bind.address:127.0.0.1}:&#36;{jboss.http.port:8080}</xsl:attribute>
+        <storage-adapter type="HAWKULAR">
```
pilhuhn: Same comment about 'xsl:choose'

PR #134, opened 2015-05-27
HAWKULAR-258 Add integration tests WildFly profile

ppalaga:

PR #148, opened 2015-05-28
HAWKULAR-279 dev profile cleanup

ppalaga:


ppalaga:
A followup of https://github.com/hawkular/hawkular/pull/134

ppalaga:
@jmazzitelli @pilhuhn could you please review?

pilhuhn:
[19:06:07]  <ppalaga>	pilhuhn: so the real and definiteve issue is: when building with -Pdev, hawkular-realm.json gets overwritten by a copy of hawkular-realm-for-dev.json
[19:06:30]  <ppalaga>	pilhuhn: hence both hawkular-realm.json and hawkular-realm-for-dev.json have the same -dev content
[19:06:40]  <ppalaga>	pilhuhn: which is unwanted IMO
[19:09:13]  <ppalaga>	pilhuhn: with PR#148, that clash will not happen

pilhuhn:
Why is that unwanted? When it is a -Pdev build, we want the -dev realm being in effect. Does not matter to me if that is via copying that over or not. 
Developers do not expect that a "dev container" (-Pdev build) can be magically turned into a non-dev one.

ppalaga:
Well, first of all please note that I stopped tending to advocate the view that devs should start WF with an explicit `-c standalone-itest.xml`. If others find it important to omit `-c standalone-itest.xml`, I really will not block that.

The present PR is just about making the current state (in which `-c standalone-itest.xml` is needed) consistent. `-c standalone-itest.xml` is really needed, because the default `standalone.xml` does not contain all the -dev stuff -- just compare `standalone.xml` with `standalone-itest.xml`. 

Overwriting `standalone.xml` with `standalone-itest.xml` when `-Pdev` is active was not a part of the present objective, because that requirement came to my consciousness just after I submitted the present PR.

I must say that I expected some of the proponents of the `standalone.xml == standalone-itest.xml <=> -Pdev` doctrine to come up with a patch to make it real. However, if you state clearly that you expect that from me in the present PR, I can do that :) Otherwise, this can be merged immediately, can it not?

ppalaga:
@pilhuhn the last commit brings back the feature that `standalone.xml` contains development settings when built with `-Pdev`

pilhuhn:
On 1 Jun 2015, at 14:55, Peter Palaga wrote:

> @pilhuhn the last commit brings back the feature that `standalone.xml` 
> contains development settings when built with `-Pdev`

@jmazzite can you please have a look?

jmazzitelli:
I'm +1 on this.

PR #149, opened 2015-05-28
add new hawkular component log categories

jmazzitelli:
set console log level appropriately
use xsl:choose, not xsl:if


jshaughn:
Although the travis job failed I can see that the work in this PR is working. The problem is elsewhere in Hawkular.  So, merging...

ppalaga:
@jshaughn this commit caused that Hawkular instance started for the end to end tests produces tons of debug output that exceeds Travis limits:
```
The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.
```

I vote for reverting this commit because no single PR will pass from now on unless somebody fixes it separately.

pilhuhn:
Just as side note, this got addressed inside #151

PR #153, opened 2015-05-29
move to agent 0.0.6

jmazzitelli:
the UI needs to know what display name to show for resources. this moves the agent to 0.0.6 which will add a name property to all resources so the UI can refer to them


mtho11:
Got the go ahead from Mazz to merge this even though its failing with some bogus Travis errors.

PR #154, opened 2015-05-29
Fixed Hawkular console version

theute:


pilhuhn:
This builds locally with that command line:

    mvn clean install -Pdev -DskipTests -Pdev

So I guess it is safe to merge anyway

PR #157, opened 2015-05-29
HAWKULAR-281: Response time: -1 when the pinger can't tell the respon…

theute:
…se time

Display "n/a" instead of "-1 ms"


pilhuhn:
lgtm - Travis failure is due to build unable to shut down a non-running server

PR #159, opened 2015-05-29
Lock down the last 2 bower packages

mtho11:
Lock down the last two hawkular-charts and hawkular-ui-services packages in preparation of stable release.


mtho11:
Merging despite bogus Travis messages.

PR #165, opened 2015-06-01
move to agent v0.0.8

jmazzitelli:
move to agent 0.0.8; provides more unique resource names for those that are the same across multiple servers


mtho11:
@jmazzitelli  can you rebase again so it can be automerged.

PR #169, opened 2015-06-01
HWKINVENT 45 - Deleting metrics before deleting the resource from inventory

Jiri-Kremser:
also when creating and deleting the metrics, now we are doing it in parallel with the `$q.all`


mtho11:
LGTM.

PR #174, opened 2015-06-02
HAWKULAR-146: New Application Server List page

ammendonca:


ui/console/src/main/scripts/plugins/metrics/plugins/metrics/ts/metricsPlugin.ts
```diff
@@ -69,6 +69,40 @@ module HawkularMetrics {
         };
     });
 
+    /**
+     * Replicates AngularJS 1.4 limitTo filter
```
mtho11: Angular 1.3 has a limitTo filter: 
https://github.com/angular/angular.js/blob/g3_v1_3/src/ng/filter/limitTo.js

ammendonca: Right, but it doesn't have the "begin" argument which is handy for this situation. We can remove this if/when we upgrade to 1.4

mtho11:
Merging. Travis failed on some bogus errors.

PR #175, opened 2015-06-02
HAWKULAR-297 It wasn't possible to add, remove and again the same URL under the sa…

Jiri-Kremser:
…me persona, because of existing alert trigger instance.


pilhuhn:
Which jira is that?

Jiri-Kremser:
none, I've just discovered that 30mins ago, I can create one..

PR #176, opened 2015-06-02
HAWKULAR-300 Fix the availability refresh button. 

mtho11:
Fix the availability refresh button(it was using  an old method name) and Make the salutation filter just upper case the first character because some names may want to preserve other case in the name.


mtho11:
Merging despite bogus C* errors on Travis.

PR #177, opened 2015-06-02
[HAWKULAR-299] A resource disappearing asynchronously while traits publisher

metlos:
tries to write to it is a valid scenario and should not result in an
unhandled exception.

### DIFF
--- a/modules/pinger/src/main/java/org/hawkular/component/pinger/Log.java
+++ b/modules/pinger/src/main/java/org/hawkular/component/pinger/Log.java
@@ -54,4 +54,7 @@ public interface Log extends BasicLogger {
             "Code %d, message= %s")
     void wNoInventoryFound(int status, String reasonPhrase);
 
+    @LogMessage(level = Logger.Level.INFO)
+    @Message(id = 5005, value = "Resource '%s' of tenant '%s' is no longer present in inventory.")
+    void iResourceNotFound(String resourceId, String tenantId);
 }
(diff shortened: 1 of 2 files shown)


ppalaga:
Why are all Inventory exceptions unchecked?

metlos:
Why are all JPA exceptions unchecked?

ppalaga:
@metlos sorry, I do neither know nor have found quickly enough why JPA exceptions are unchecked. Could you please explain as shortly as possible?

ppalaga:
@metlos could you please {{s/HAWKULAR-255/HAWKULAR-299/}} in the commit message?

metlos:
Well, that is quite a complicate question to answer and is of course partially a matter of taste.

But the reasoning behind not using checked exceptions is that rarely the caller has any reasonable way of handling the exception. So by using a documented unchecked exception you give the caller the OPTION to handle the exception but you do not force them to do so if they have no way of sensibly handling it.

http://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html

IMHO, `EntityNotFoundException` is one that the caller cannot be expected to reasonably recover from - they asked for something which isn't there. A typical catch block for such exception would be to merely rethrow it, so it doesn't make sense to make a checked exception.

metlos:
Corrected the JIRA number.

ppalaga:
@metlos thanks for the explanation. For me as an API user, it is a disadvantage to have to search for unchecked exceptions that I could want to handle. Moreover, there is a very good chance that I'll miss future changes in thrown exceptions. With checked exceptions the compiler would fail or at least warn me on any change.

ppalaga:
I find the change OK to merge once the latest version passes CI.

metlos:
CI passed.

PR #190, opened 2015-06-04
HAWKULAR-272: Fixed alert settings time-related bugs

vrockai:


ui/console/src/main/scripts/plugins/metrics/plugins/metrics/ts/metricsAlerts.ts
```diff
@@ -209,14 +214,26 @@ module HawkularMetrics {
       this.$log.debug('this.metricId', this.metricId);
     }
 
+    // Get the most meaningful time unit (so that time value is not a very long fraction).
+    private getTimeUnit(timeValue: number): number {
```
mtho11: I think some of this time unit conversion stuff can be reusable and should be put in a service.

PR #191, opened 2015-06-05
HAWKULAR-309 Improve debug output in Pinger and Availability Creator

ppalaga:


ppalaga:
@metlos, could you please review?

modules/pinger/src/main/java/org/hawkular/component/pinger/PingManager.java
```diff
@@ -171,6 +178,7 @@ private void doThePing(Set<PingDestination> destinations) {
             final long now = System.currentTimeMillis();
             PingStatus ps = PingStatus.timeout(destination, now, TIMEOUT_MILLIS);
             results.add(ps);
+            Log.LOG.debugf("Timeouted: %s", ps.getDuration());
```
pilhuhn: Timed out (?)

ppalaga: +1, even `ps.getDuration()` is wrong. Should be `destination.getUrl()`.

PR #192, opened 2015-06-05
HAWKULAR-323 Add log level system variables

ppalaga:


ppalaga:
[Don't Merge Yet]

jmazzitelli:
I think this is fine. I don't consider this polluting standalone.xml - we are adding these categories already - its just the value we are setting it to can be overridden with runtime system properties. I am +1 on this. anyone else have opinions?

tsegismont:
Looks good to me as well. The system property switch is nice as you can use it for a one time debug session, whereas changes with the CLI are persisted and need to be rolled back.

ppalaga:
There seems to be enough consensus in this matter. @pilhuhn could you please merge once the last rebase passes CI?

PR #194, opened 2015-06-08
HAWKULAR-329: 'miliseconds' spelled wrong

theute:
I didn't check at all the code and if this typo needs a fix in other components.


theute:
@vrockai could you review this ? Should be fine unless there are magic tricks in other places.

vrockai:
@theute no magic tricks in this case :) looks fine to me

theute:
:+1:

PR #197, opened 2015-06-08
move to agent .10

jmazzitelli:
HAWKULAR-321 this provides hostname as a resource configuration property on the top level server resource


mtho11:
We can use this now!

PR #203, opened 2015-06-11
HAWKULAR-337 Handle the race condition between WildFly Agent and End to

ppalaga:
End Tests


ppalaga:
@metlos could you please review?

metlos:
lgtm :+1:

PR #214, opened 2015-06-16
HAWKULAR-350 Do not limit the level of CONSOLE logger for kettle.build.type=dev

ppalaga:


ppalaga:
@jmazzitelli could you please ack 5f22d38 ? It is a small detail, IMO fully in intentions of how we handled the `standalone.xml` files so far. Please have look into https://issues.jboss.org/browse/HAWKULAR-350 for more info.

ppalaga:
@jmazzitelli could you please ack esp. 01a76cd ? It is a small detail, IMO fully in intentions of how we handled the standalone.xml files so far. Please have look into HAWKULAR-350https://issues.jboss.org/browse/HAWKULAR-350 for more info. 
Note that this PR is against inventory-next branch. It is because I needed this stuff for my further work in this branch.

jmazzitelli:
If I read this right, if this is a dev build, this will strip out the log level setting completely from the default standalone.xml. If it is a production build, it keeps the INFO log level that comes with the default standalone.xml. Correct? If so, I'm OK with this. (I am assuming the log level setting is optional - without it, I assume the console will pass through everything).

ppalaga:
> If I read this right, if this is a dev build, this will strip out the log level setting completely from the default standalone.xml. If it is a production build, it keeps the INFO log level that comes with the default standalone.xml. Correct?

Right

> If so, I'm OK with this. (I am assuming the log level setting is optional - without it, I assume the console will pass through everything).

Right

ppalaga:
BTW the last commit depends on https://github.com/hawkular/hawkular-inventory/pull/87 in inventory which is not in master yet.

dist/src/main/resources/wildfly/patches/standalone.xsl
```diff
@@ -140,10 +140,7 @@
   <!-- set the console log level -->
   <xsl:template match="logging:console-handler[@name='CONSOLE']/logging:level"
                 xmlns:logging="urn:jboss:domain:logging:2.0">
-    <xsl:choose>
-      <xsl:when test="$kettle.build.type='dev'"><level name="DEBUG"/></xsl:when>
-      <xsl:otherwise><level name="INFO"/></xsl:otherwise>
-    </xsl:choose>
+    <xsl:if test="$kettle.build.type='production'"><level name="INFO"/></xsl:if>
```
jpkrohling: What's the end effect of this? Without testing, I'd say that if dev is being used, it assumes INFO, as it's the default on the root logger. Or is it specified somewhere else that this is TRACE (or whatever is lower)?

modules/end-to-end-test/src/test/groovy/org/hawkular/inventory/rest/test/InventoryITest.groovy
```diff
@@ -157,14 +185,14 @@ class InventoryITest extends AbstractTestBase {
         Collections.reverse(entries)
         for (Map.Entry en : entries) {
             String path = en.getKey();
-            boolean validate = en.getValue();
+            String getValidationPath = en.getValue();
```
jpkrohling: I know this is a Groovy test, but isn't this name too close to a getter? :)

dist/src/main/resources/wildfly/patches/standalone.xsl
jpkrohling: Just noted the previous comments about this part, so, please disregard this one :)

jpkrohling:
Looks good to me, but I cannot judge the accuracy of the tests, nor if they are sufficient.

modules/end-to-end-test/src/test/groovy/org/hawkular/inventory/rest/test/InventoryITest.groovy
ppalaga: Yes, the name may resemble a getter method. Clearly, the meaning of the name is "the path to use when checking that the deletion was successful" - do you have any better idea?

jpkrohling: validationPathGET?

PR #223, opened 2015-06-19
Upgrade to hawkular-parent 16 - upgarde to WildFly 9

ppalaga:


theute:
There is much more that that to do in Hawkular:
   - release and  upgrade Bus
   - release and upgrade Agent
   - Makes changes in Hawkular: https://github.com/hawkular/hawkular/tree/mazz/wf9

I think @jmazzitelli said he would take care of that later today

ppalaga:
Already in master

PR #226, opened 2015-06-19
HAWKULAR-348 : Display n/a for response time aggregates when the values are < 0

ammendonca:


ui/console/src/main/scripts/plugins/metrics/plugins/metrics/html/response-time.html
```diff
@@ -57,7 +60,7 @@ <h3 class="pull-left">Metrics</h3>
       </div>
     </div>
 
-    <div class="panel panel-default hk-graph">
+    <div class="panel panel-default hk-graph" ng-if="vm.average >= 0">
```
mtho11: I think we want to use ng-show/ng-hide instead of ng-if as it removes/adds DOM which is expensive.

ammendonca: Disagree, the fact that it removes/adds DOM was the reason why I decided to use ng-if, because this is a condition that will not vary constantly (worst case, it changes once every hour). Now, if the chart is just hidden, all evaluations are performed but the chart is just not visible.. if it is removed from the DOM all the processing to calculate the data points and to render them it is skipped as well, with a significant gain in performance at every 20sec refresh and as well at total document size, which is about 16kb with ng-if versus about 70kb if ng-show is used (when the chart is not to be shown).
The usage of ng-if should be weighted, but this is the kind of scenario it exists for.

mtho11: Yeah, I'll buy that. Just have to ask why ng-if and you have a legitimate reasoning behind use of it :+1:

PR #231, opened 2015-06-22
HWKINVENT-18 REST tests for inventory relationships

Jiri-Kremser:


modules/end-to-end-test/src/test/groovy/org/hawkular/inventory/rest/test/InventoryITest.groovy
```diff
@@ -181,9 +185,22 @@ class InventoryITest extends AbstractTestBase {
         pathsToDelete.put("$path/$responseStatusCodeMetricId", "$path/$responseStatusCodeMetricId")
 
         /* add a feed */
-        response = postDeletable(path: "/hawkular/inventory/$environmentId/feeds", body: [ id : feedId ])
+        response = postDeletable(path: "$environmentId/feeds", body: [id: feedId])
+        assertEquals(201, response.status)
+        assertEquals(baseURI + "$inventoryEndpoint/$environmentId/feeds/$feedId", response.headers.Location)
+
+        /* add a custom relationship, no need to clean up, it'll be deleted together with the resources */
+        def relation = [id        : 42, // it's ignored anyway
```
metlos: This is confusing - I didn't notice it in the original PR but do users really have to supply an ID for the relationship if it is going to get ignored? What about using a blueprint type that wouldn't have properties that users don't need to provide?

metlos:
I've created https://issues.jboss.org/browse/HWKINVENT-75 to capture the commented issue. Otherwise this looks good. Thanks!

PR #232, opened 2015-06-22
Bump hawkular-charts version to include the change from HAWKULAR-213

pilhuhn:


mtho11:
The bower.json rule:  "hawkular-charts": "^0.3.5" means that it will pull in all patch version > 0.3.5. So really isn,t necessary. When I build Hawkular locally I get the 0.3.9 hawkular-charts version.

pilhuhn:
On 22 Jun 2015, at 18:20, Mike Thompson wrote:

> The bower.json rule:  "hawkular-charts": "^0.3.5" means that it will 
> pull in all patch version > 0.3.5. So really isn,t necessary. When I 
> build Hawkular locally I get the 0.3.9 hawkular-charts version.

I re-built master several times and it did not work.
Only when I explicitly requested 0.3.9 as in the patch.

mtho11:
I'll merge for now since it can't hurt and evidently can help. But this is strange behaviour.

mtho11:
A fun way to plan around with these rules visually: http://semver.npmjs.com/

PR #237, opened 2015-06-23
HAWKULAR-376 - Added redirect URIs and web origins for accounts backend client.

jpkrohling:


jpkrohling:
@Jiri-Kremser  , can you please check this?

dist/src/main/resources/configuration/hawkular-realm-for-dev.json
```diff
@@ -86,6 +86,8 @@
       "enabled": true,
       "bearerOnly" : false,
       "publicClient": false,
+      "redirectUris": ["/*", "http://localhost:2772/*"],
```
Jiri-Kremser: are you sure about the port? Why not 8080?

dist/src/main/resources/configuration/hawkular-realm-for-dev.json
```diff
@@ -86,6 +86,8 @@
...
+      "webOrigins": ["http://localhost:2772"],
```
Jiri-Kremser: Btw. there is also https://issues.jboss.org/browse/HAWKULAR-372 I guess this is the right place to add also following origins: http://pastebin.test.redhat.com/292036, wdyt? I can create a PR for it not to complicate things and solve them separately.

dist/src/main/resources/configuration/hawkular-realm-for-dev.json
jpkrohling: It's relative to the main URL, so, it will allow the same protocol, hostname and port as the auth server (ie: Keycloak).

dist/src/main/resources/configuration/hawkular-realm-for-dev.json
jpkrohling: I would prefer to track those differently.

dist/src/main/resources/configuration/hawkular-realm-for-dev.json
jpkrohling: By the way: please test it and see if it solves your original issue.

Jiri-Kremser: it works as expected, merging..

PR #239, opened 2015-06-23
HAWKULAR-372 - adding special origins for PhantomJS browser

Jiri-Kremser:
ui-services REST tests need it to work


Jiri-Kremser:
@jpkrohling could you please review?

jpkrohling:
You don't need this for the main realm file, shipped with the "production" profile, right? Only with the "dev" version, I assume. If so, it looks good to me.

Jiri-Kremser:
yes, we run the tests only against the dev instance, because the jdoe is assumed

jpkrohling:
Looks good to me :)

PR #242, opened 2015-06-23
Set standalone.xml default hawkular.metrics.waitForService=true.  Thi…

jshaughn:
…s stops

dependent deployments from deploying before metrics is ready and the schema
is installed.  Also, add embedded C* in dep mgmt.


ppalaga:
@jshaughn `hawkular-next` was merged to `master` and it looks like this PR missed the train, which is my fault probably. - Sorry for that. In this situation, I can do nothing else then to close this PR.

Could you please 
(1) Rebase your branch on top of master and 
(2) Send a new PR

PR #251, opened 2015-06-25
Bump the hawkular-charts version to 0.3.13

mtho11:


mtho11:
Will move to fixed bower versions of hawkular-ui-services and hawkular-charts, now that we are close to release.

PR #271, opened 2015-06-29
HAWKULAR-341 URLs with 2 x-powered-by header fail

ppalaga:


pilhuhn:
As you are separating the values by comma, wouldn't it be easier to skip the multi-value case and have a map<header,String> where you check if get(header) is not null and if so you just append ", " + other value?  Or do you fear duplicates here?

ppalaga:
@pilhuhn I guess, duplicates are not that frequent in the wild web (although this solves the problem well, when they occur). My main motivation for the present approach has to do with how we look at the identity of the values:

(1) Note that the present approach sorts the multiple values alphabetically. Having some kind of canonical order should make aggregation over x-powered-by more reliable. Let's say we want to count instances by used technology. With sorted values, both `"ARR/2.5, ASP.NET"` and `"ASP.NET, ARR/2.5"` will be counted together, which is good.

(2) The same holds for the detection of changes. I have no idea how reliable is it to assume that the servers are sending the headers always in the same header. I doubt that there some spec prescribing them to send the headers in the same order for two identical (or even in some sense similar) requests. I think it is theoretically possible that a server sends `"ARR/2.5, ASP.NET"` once and `"ASP.NET, ARR/2.5"` other time. With the present approach, there is no change for the inventory, which is good IMO again.

Does that sound acceptable?

pilhuhn:
Sure, was just asking :)

PR #273, opened 2015-06-29
Hwkinvent 78

ppalaga:


ppalaga:
@metlos could you please review and eventually merge?

ppalaga:
Failed because it depends on https://github.com/hawkular/hawkular-inventory/pull/101

metlos:
Shouldn't the last commit upgrade inventory to 0.1.0.Final?

metlos:
Oh, it does :) Only the commit message is wrong.

ppalaga:
@metlos fixed the last message. Sorry.

metlos:
lgtm. This needs to go inline with a new release of agent that will adapt to the same changes accordingly.

PR #274, opened 2015-06-29
JMAN4-60 add some undertow metrics

jmazzitelli:


mtho11:
LGTM. We can use these.

PR #284, opened 2015-06-30
Update dist to use Inventory 0.1.0.Alpha3.

metlos:
**This is different from https://github.com/hawkular/hawkular/tree/dev/inventory-0.1.0.Alpha3**.

On top of the tested `0.1.0.Alpha2` which is currently in master, `0.1.0.Alpha3` contains only a fix for bug https://issues.jboss.org/browse/HWKINVENT-54 and therefore gets rid of the ugly error where agent is not able to fully sync its resource types with the server.

No other changes are present and most notably, no UI changes are necessary for this to work in Hawkular 1.0.0.Alpha2.


pilhuhn:
Very good. I was thinking of going that route as well


Am 30. Juni 2015 23:25:15 schrieb Lukas Krejci <notifications@github.com>:

PR #285, opened 2015-07-01
HAWKULAR-219 : Integrate Datasource charts

ammendonca:


ui/console/src/main/scripts/plugins/metrics/plugins/metrics/html/app-details/detail-datasources.html
```diff
@@ -55,33 +55,55 @@ <h3 class="pull-left">Metrics</h3>
...
-            <h4>Responsiveness<!-- <img src="img/legend-responsiveness.png" alt="legend-responsiveness" height="12"> --></h4>
+        </div>
+      </div>
+
+      <div class="panel panel-default hk-graph" ng-if="vm.chartRespData[ds.id][1]">
+        <h2>Responsiveness</h2>
+        <div class="metrics-chart">
+          <hawkular-chart
+              multi-data="{{vm.chartRespData[ds.id]}}"
+              chart-type="multiline"
```
pilhuhn: This (and above) looks pretty odd -- didn't we say we want the Gauges type.
And a unit of bytes for number of available datasources looks wrong too.
Or is this just a placeholder?

ammendonca: Don't recall we moving to gauge type for these charts, would they make sense for this data? This is for the expanded "View Metrics" showing pool availability (available/inuse/timedout connections) and responsiveness (conn wait and creation time), mocks use muiltiline: https://issues.jboss.org/secure/attachment/12390611/Datasources3-expanded2.jpg

I've used two different charts as our multiline is not ready for double y-axis units, so it would look odd as units scale differ a lot. The unit type "bytes" was forgotten, had it correct but undo-ed some changes to fix a problem and forgot about it. will fix.

We don't have the Gauge chart type ready for the connections used.

pilhuhn: Ah sorry. I got that wrong and confused it with the gauge type in the main box. My bad.

PR #286, opened 2015-07-01
HAWKULAR-366 : Fix availability info for URL list

ammendonca:
The fix reverses the results array, but metrics should return it in reverse order (more recent first)

### DIFF
--- a/ui/console/src/main/scripts/plugins/metrics/plugins/metrics/ts/addUrlPage.ts
+++ b/ui/console/src/main/scripts/plugins/metrics/plugins/metrics/ts/addUrlPage.ts
@@ -194,7 +194,8 @@ module HawkularMetrics {
           promises.push(this.HawkularMetric.AvailabilityMetricData(tenantId).query({
             availabilityId: res.id, distinct: true,
             start: 1, end: moment().valueOf()}, (resource) => {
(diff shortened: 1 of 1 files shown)


ui/console/src/main/scripts/plugins/metrics/plugins/metrics/ts/addUrlPage.ts
```diff
@@ -194,7 +194,8 @@ module HawkularMetrics {
           promises.push(this.HawkularMetric.AvailabilityMetricData(tenantId).query({
             availabilityId: res.id, distinct: true,
             start: 1, end: moment().valueOf()}, (resource) => {
-            res['isUp'] = (resource[0] && resource[0].value === 'up');
+              resource.reverse(); // FIXME: HAWKULAR-366
```
pilhuhn: This is just reversing the list not doing a real sort?  Wouldn't a real sort be better, as this is as prone to changes in the server as before? But then it is probably good enough for now.

ammendonca: IMO, is better that it fails when server changes so we remember to remove the unnecessary operation. Plus sorting would be heavier, for something we know that reverse is what we want.

PR #289, opened 2015-07-01
HAWKULAR-411 - max heap should be a metric

jmazzitelli:


dist/src/main/resources/wildfly/patches/standalone.xsl
```diff
@@ -377,6 +377,11 @@
                       timeUnits="minutes"
                       path="/core-service=platform-mbean/type=memory"
                       attribute="heap-memory-usage#committed" />
+          <metric-dmr name="Heap Max"
```
pilhuhn: Sure with "Heap Max" wrt previous "Max Heap" ?

jmazzitelli: I only named it that to be consistent with the names of the other similar metrics:
   "Heap Used"
   "Heap Committed"

PR #300, opened 2015-07-07
Delete old Alerts plugin.

mtho11:
It doesnt seem like this will ever be used and it represents technical debt whenever we do upgrades or anything else global.
Maybe it should just be moved somewhere else? Anyway, it needs to be maintained to keep up with all of  the changes.

I know @lucasponce  put a lot of work into this and feel bad about deleting (my metrics-console and metrics-explorer have also gone to code heaven).

CC: @jshaughn @pilhuhn


mtho11:
Let's wait to merge until after Lucas has had a chance to respond.

jshaughn:
Is this the sample UI Lucas wrote for you guys as an example of the API?  If so it was never anticipated for it to live forever, I think he expects and will be fine with the removal. It's now out of date.

mtho11:
Yes, @jshaughn it is that sample alerts UI.

lucasponce:
Hi @mtho11, no problem from my side.
The goal of this code was to help in some way to show you how to call the first API we created for alerts.
I'm happy if it helped so please clean the code as I can see that it has no sense to maintain it.
Thanks !

mtho11:
Merged #300

PR #301, opened 2015-07-07
Delete old Inventory plugin.

mtho11:
This is the old inventory plugin from many months ago. Since then, inventory has changed greatly. Not sure if this is still needed as it represents technical debt.

CC: @pilhuhn @ammendonca


mtho11:
@pilhuhn Please merge if you agree.

mtho11:
Merged #301.

PR #312, opened 2015-07-10
Removed extra dependency sets for Accounts KC event listeners.

jpkrohling:
@pilhuhn  , this PR should solve the issue you reported via email :)

### DIFF
--- a/dist/assembly.xml
+++ b/dist/assembly.xml
@@ -159,29 +159,6 @@
       <useTransitiveFiltering>false</useTransitiveFiltering>
     </dependencySet>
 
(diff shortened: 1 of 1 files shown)


pilhuhn:
So you say that they are not needed? Or should they end up in dist?

jpkrohling:
They are duplicated. The PR that added those got a comment from @jmazzitelli , to add those listeners as ZIP jboss modules instead of assembling the modules at hawkular itself. So, the PR that got merged had those two on the assembly.xml , and this PR removes the JAR versions, while keeping the ZIP versions.

PR #314, opened 2015-07-10
Add .editorconfig file for multiple editors to honor spacing/tabs. Se…

mtho11:
…e http://editorconfig.org

### DIFF
--- a/console/src/main/scripts/.editorconfig
+++ b/console/src/main/scripts/.editorconfig
@@ -1,0 +1,14 @@
+# EditorConfig: http://EditorConfig.org
+
+# top-most EditorConfig file
+root= true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.{js,ts,html,less,css,json}]
+indent_style = space
+indent_size = 2
--- a/console/src/main/scripts/plugins/metrics/ts/metricsResponsePage.ts
+++ b/console/src/main/scripts/plugins/metrics/ts/metricsResponsePage.ts
@@ -54,8 +54,7 @@ module HawkularMetrics {
    */
   export class MetricsViewController {
     /// for minification only
-    public static  $inject = ['$scope', '$rootScope', '$interval', '$log', 'HawkularMetric', 'HawkularAlert',
-      '$routeParams','HawkularAlertsManager' ,'HawkularErrorManager', 'AlertService'];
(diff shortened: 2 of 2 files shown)


ppalaga:
I am all for having .editorconfig that will be observed across editors, IDEs and other tools such as Maven. As it stands ATM, I do not see a Maven plugin that would enforce .editorconfig and the Eclipse support seems to be in early phases too. 

@mtho11 could you please list editors and tools that work for you supporting .editorconfig?

Other than that, the present PR does not compile:
```
[INFO] (max-line-length) metrics/ts/metricsResponsePage.ts[57, 1]: exceeds maximum line length of 120
[ERROR] 
[ERROR] stream.js:94
[ERROR]       throw er; // Unhandled stream error in pipe.
[ERROR]             ^
[ERROR] Error: Failed to lint: metrics/ts/metricsResponsePage.ts[57, 1]: exceeds maximum line length of 120.
```

vrockai:
@ppalaga they have the list (including an eclipse plugin) on their page:
http://editorconfig.org/#download

ppalaga:
Ah, noticed now that the `.editorconfig` file is in `console/src/main/scripts` and it is supposed to be used by gulp or whatever javascript tools. Missing Maven plugin does not hurt there at all.

ppalaga:
@vrockai I saw that list, but I am quite skeptical about what such a list can mean. I wanted to hear from a team mate what works for him in his real daily use cases.

vrockai:
@ppalaga , it's intended for auto-configuration of an editor (editor should pick this file automatically), I'm not aware of any other intentions (like maven or gulp). AFAIK it was added mainly to prevent different indentations in the code by different developers.

ppalaga:
@vrockai I agrees, but `.editorconfig` makes sense only if editors used throughout the team support it. Which editors are both (i) used by you and (ii) supporting it?

vrockai:
@ppalaga , we all use (supported) Webstorm in the UI sub-team. I, personally, have never used any other editor for work than listed in the editorconfig supported list.

ppalaga:
Note that after fixing `metricsResponsePage.ts`, the build fails on `.editorconfig` not being recognized by maven license plugin, which I am about to address in the next HK Parent version.

ppalaga:
Do not merge to master before hawkular-parent 18 is released.

mtho11:
This is just a nice to have feature. But does seem to be the way to support some basic code styling in the  actual editors (where the code is written instead of external tooling). And the editor support is very extensive as shown by @vrockai's link above.

ppalaga:
Replaced by #333

PR #319, opened 2015-07-14
HAWKULAR-444 - Clear token on timeout, so that background requests are blocked

jpkrohling:


jpkrohling:
@mtho11 could you please review this one? Note that I added a new dependency on keycloak (which was a transitive dependency, through hawtio-oauth)

jpkrohling:
It seems that @mtho11 is on PTO, so, could you please review this one, @vrockai ?

vrockai:
looks good to me :thumbsup:

mtho11:
Thanks vrockai!

PR #322, opened 2015-07-15
Console: using tsd for typescript defs

vrockai:


console/src/main/scripts/gulpfile.js
```diff
@@ -152,11 +160,11 @@ var gulpTsc = function(done) {
     });
 };
 
-gulp.task('tsc', ['clean-defs'], function (done) {
+gulp.task('tsc', ['clean-defs', 'tsd'], function (done) {
   gulpTsc(done);
 });
 
-gulp.task('tsc-live', ['copy-sources','clean-defs'], function (done) {
+gulp.task('tsc-live', ['copy-sources','clean-defs', 'tsd'], function (done) {
```
mtho11: I'm not sure that this needs to run with every change. The type definition files shouldnt change much and it slows the compiling down to have to hit those urls and pull in that data. Right now there is only one type def file and I can see the slowdown. With 10+ type def files its a real tax. Probably even too much for Travis builds (especially given the static nature of those files it isn't necessary). 
This is a great task to have, just think it should be run manually (and type defs checked in) not automatically.

console/src/main/scripts/plugins/directives/vendor/moment/moment.d.ts
```diff
@@ -1,472 +0,0 @@
-// Type definitions for Moment.js 2.8.0
```
mtho11: We still need to add the rest of the d.ts files for all the other libraries in the project. Moment.d.ts is the only type def file.

console/src/main/scripts/gulpfile.js
vrockai: Ouch, I have expected this to not be executed if there are no changes (the same as npm or bower). If that's not the case right now, it should be definitely fixed somehow.

console/src/main/scripts/plugins/directives/vendor/moment/moment.d.ts
vrockai: The problem is in the hawt.io dependencies. The hawt.io packages contain their own definitions, which cause conflicts.

PR #323, opened 2015-07-15
Fix the line length in metricsResponsePage.ts and add lincense header to

ppalaga:
.editorconfig


ppalaga:
@vrockai could you please review and eventually merge to `editorconfig` branch?

Please note that `editorconfig` branch is waiting for the next `hawkular-parent` release so that `.editorconfig` file type is properly recognized.

vrockai:
I'm don't have merge permissions, but the PR looks good to me.

PR #324, opened 2015-07-15
HAWKULAR-415 & HAWKULAR-347 : Improve Availability Info

ammendonca:


console/src/main/scripts/plugins/metrics/html/availability.html
```diff
@@ -24,11 +24,22 @@ <h3 class="pull-left">Availability Status</h3>
...
         <div class="col-sm-3 hk-summary-item">
-          <span class="hk-data">{{vm.downtimeDuration / 1000 | number:0}} s</span>
+          <!--
+          <span class="hk-data" ng-if="vm.downtimeDuration < 60000">{{ vm.downtimeDuration / 1000 | nduration:'s\' seconds\'' }}</span>
+          <span class="hk-data" ng-if="vm.downtimeDuration >= 60000 && vm.downtimeDuration < 7200000">{{ vm.downtimeDuration | nduration:'m\' minutes and \'ss\' seconds\'' }}</span>
+          <span class="hk-data" ng-if="vm.downtimeDuration >= 7200000 && vm.downtimeDuration < 172800000">{{ vm.downtimeDuration | nduration:'h\' hours and \'m\' minutes\'' }}</span>
+          <span class="hk-data" ng-if="vm.downtimeDuration >= 172800000">{{ vm.downtimeDuration | nduration:'d\' days and \'h\' hours\'' }}</span>
+          -->
+          <span class="hk-data" ng-hide="vm.downtimeDuration">Always Up</span>
+          <span class="hk-data" ng-if="vm.downtimeDuration && vm.downtimeDuration < 60000" tooltip-trigger tooltip-placement="top" tooltip-popup-delay="1500" tooltip="{{ vm.downtimeDuration | duration:'ss\'s\'' }}">{{ vm.downtimeDuration | duration:'s\' seconds\'' }}</span>
```
mtho11: Could we move the logic out to the controller/service that returns a downtimeDuration with a 'value' and 'units' fields?

ammendonca: Indeed, I've started to make some changes to something like that, but it would require several changes (create the proper types, adapt everything) so I left if for a future commit, so it won't "hide" the actual changes to fix the issues.

mtho11: Ok, I'll merge this now to get the fix in and you can add fixes later.

PR #325, opened 2015-07-16
HWKINVENT-69 | inventory ping integration test

pavolloffay:


ppalaga:
@pavolloffay should this not better go to hawkular:dev/inventory-0.2.0.Alpha1 branch?

pavolloffay:
yes, but it should also work on master. 

I will put it into 0.2.0.Alpha1,

ppalaga:
Yes, I agree this would also work on master, but because we do a lot of work in dev/inventory-0.2.0.Alpha1, it is just safer to send all changes there just to avoid conflicts.

pavolloffay:
See https://github.com/hawkular/hawkular/pull/334

PR #332, opened 2015-07-17
[HWKINVENT-91] fixed

jkandasa:
https://issues.jboss.org/browse/HWKINVENT-91


mtho11:
Closing as -Xmx is the correct max heap value here.

PR #333, opened 2015-07-17
Add .editorconfig file for multiple editors to honor spacing/tabs

ppalaga:


ppalaga:
This is basically #314 targeted at dev/inventory-0.2.0.Alpha1 branch where we already have Parent 18

PR #338, opened 2015-07-20
Integration latest of everything

jmazzitelli:


Jiri-Kremser:
travis fails during gulp build because of 
`unexpected status code: 403 on: `https://api.github.com/repos/borisyankov/DefinitelyTyped/branches/master

we should consider using the oauth token because github has limits and travis infrastructure must be spamming them (website repo has similar issue)

 https://github.com/DefinitelyTyped/tsd/issues/74

Jiri-Kremser:
as per discussion on IRC, let's continue in master

PR #341, opened 2015-07-20
Fixing the gulp build (Invalid use of 'eval')

Jiri-Kremser:
In TypeScript it's not possible to override/shadow the eval function in the strict mode


mtho11:
Merged #341

PR #342, opened 2015-07-20
Dev profile will not create the Hawkular archives.

Jiri-Kremser:
EDIT: The `dev` profile will not create the zip and gz archives, if you need to have the old behavior use the `dozip` profile.

````bash
$cd dist
$mvn clean install -DskipTests -Pdozip
Total time: 41.865 s

$mvn clean install -DskipTests -Pdev
Total time: 22.411 s
````
=> 19.5 seconds saved

Here is what you can do in 19 seconds:
* https://www.youtube.com/watch?v=Yn9DpFY1X7o
* http://www.dailymail.co.uk/sport/olympics/article-3044119/Usain-Bolt-want-run-200-metres-19-seconds-fully-fit-impossible-beat-me.html
* https://www.youtube.com/watch?v=h2n9vPuzvj4

:smiley:

### DIFF
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -420,6 +420,24 @@
             <configuration>
               <formats>
                 <format>dir</format>
+              </formats>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <id>dozip</id>
+      <properties>
+        <kettle.build.type>dev</kettle.build.type>
+      </properties>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-assembly-plugin</artifactId>
+            <configuration>
+              <formats>
+                <format>dir</format>
                 <format>zip</format>
                 <format>tar.gz</format>
               </formats>
@@ -428,5 +446,6 @@
         </plugins>
       </build>
     </profile>
+
   </profiles>
 </project>


pilhuhn:
I like the idea per se, but not the addition of a new profile.

I think it should be the other way around. -Pdev should skip creating the archives and if you insist on having archives in dev-mode, then add the other profile or -Ddozip 

A non-dev build always creates the archives, so we don't lose that functionality for releases etc.

Jiri-Kremser:
I agree, but I don't know who depends on it. It's possible that someone builds it with `-Pdev` and do something with the zip, I could imagine QE is doing this in foreman. I wanted to be minimally invasive and don't break anything.

Jiri-Kremser:
> I think it should be the other way around. -Pdev should skip creating the archives

Done. Hopefully, it won't break anything, I can send it to the mailing list that people can use the dozip profile if they need the archive with jdoe

pilhuhn:
On 21 Jul 2015, at 11:29, Jirka Kremser wrote:

> I agree, but I don't know who depends on it. It's possible that 
> someone builds it with `-Pdev` and do something with the zip, I could 
> imagine QE is doing this in foreman. I wanted to be minimally invasive 
> and don't break anything.

Then ask QE and warn them that this change may be coming.
We had too many profiles in the past, which was pretty confusing
to many members of the community.

Jiri-Kremser:
travis fails during gulp build because of 
`unexpected status code: 403 on: `https://api.github.com/repos/borisyankov/DefinitelyTyped/branches/master

we should consider using the oauth token because github has limits and travis infrastructure must be spamming them (website repo has similar issue)

mtho11:
I like this!
Too many profiles are a concern, but this is a real time savings.

mtho11:
Merged #342

PR #343, opened 2015-07-21
Updated gmaven maven plugin

jpkrohling:


jpkrohling:
It seems that 1.3 has some problem on central.

PR #346, opened 2015-07-21
HAWKULAR-473 Make tsd a manual task to update the type definition fil…

mtho11:
…es (typings/*.d.ts), instead of automatically downloading every build. These files only chnage with library version changes.


mtho11:
Merged #346

PR #349, opened 2015-07-22
HAWKULAR-398 add accumulated GC metric

jmazzitelli:


mtho11:
@ammendonca this is for you!

PR #363, opened 2015-07-28
Fix per server deployments filtering

ppalaga:


ppalaga:
@mtho11 could you please review?

mtho11:
@ppalaga LGTM :+1:

PR #366, opened 2015-07-29
HAWKULAR-492 Hawkular UI Url detail Alerts Tab Nothing shows up

mtho11:
Overzealous refactoring without using 'this'.


mtho11:
Merged #366

PR #370, opened 2015-07-29
turn three metrics into counters

jmazzitelli:
there are more that we need to designate counter, but rather than flip them all right before release, I'll just do these three

h-434 is the jira where we need to go through and mark all the counters that should be counters


jmazzitelli:
amendonca gave his seal of approval

PR #377, opened 2015-07-29
Do not try be cool using angular.forEach with this in the handler

ppalaga:


ppalaga:
@mtho11 could you please review and merge?

mtho11:
Merged #377

PR #387, opened 2015-07-31
HAWKULAR-499 First pass at an Angular Style Guide

mtho11:
This is a first pass at a best practices for Hawkular UI style guide. It was mainly copied from Todd Motto and altered for our use cases (like Typescript instead of javascript). I think it summarizes some very important Angular principles in one place. 
With more people on the team authoring in Angular its is important to have this.


mtho11:
I realize this should be in asciidoc format. But most of this was copied from an existing MarkDown document. So for now...

mtho11:
@ppalaga Any reason why this would fail checkstyle with: 
[WARNING] Unknown file extension: /home/travis/build/hawkular/hawkular/angular-style-guide.adoc

ppalaga:
@mtho11 `.adoc` is indeed unknown to license plugin. It has not shown up so far because `README.*` pattern is ignored. `angular-style-guide.adoc` is probably the first non-README adoc file.

As a solution, (a) adding the whole `*.adoc` to excludes of license plugin in HK parent is probably better than (b) defining the license hearder style and adding licenses?

PR #388, opened 2015-08-03
Update the Dockerfile to Alpha4 and add more documentation

pilhuhn:
More docs.

Part of that could be replicated/moved into the Web site I guess.

### DIFF
--- a/dist/docker-compose.yml
+++ b/dist/docker-compose.yml
@@ -1,6 +1,25 @@
+#
+# Copyright 2015 Red Hat, Inc. and/or its affiliates
+# and other contributors as indicated by the @author tags.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


pilhuhn:
@vnugent If you have a better idea on how to (not) encode the hawkular version in the zip name, then let me know.

pilhuhn:
Not sure why it is failing, but it looks like

Jiri-Kremser:
> Not sure why it is failing, but it looks like some ...

nope, it's failing here https://travis-ci.org/hawkular/hawkular/builds/73890669#L1151
Don't ask me, why it's displayed only as a warning, if it fails the build

Perhaps, we should consider having some fast fail behavior in the mvn build, because it's hard to find the real issue..

Jiri-Kremser:
>@vnugent I If you have a better idea on how to (not) encode the hawkular version in the zip name, then let me know.

what about:
```bash
/opt/hawkular-*-SNAPSHOT/bin/add-user.sh hawkularadmin hawkularadmin --silent
...
CMD ["/opt/hawkular-*-SNAPSHOT/bin/standalone.sh","-b","0.0.0.0","-bmanagement","0.0.0.0","-Djboss.server.data.dir=/opt/data"]

```

or 
```bash
ENV version `curl -Ls https://raw.githubusercontent.com/hawkular/hawkular/master/pom.xml | grep "^  <version>" | sed -e "s;  <version>\([0-9a-zA-Z\.\-]*\)</version>;\1;"`
do something with $version
```
?

The second approach is more risky because GitHub sometimes denies the requests because of their REST API limits.

vnugent:
In the Hawkular ci project I pulled mvn logic out of Dockerfile and put it in a bash script

https://github.com/RedHatQE/hawkular-docker/blob/master/build.sh

PR #389, opened 2015-08-03
Adding two new handy parameters for hawkular

Jiri-Kremser:
Two things:
* adding a new logging category for rest requests in inventory 
`-Dhawkular.log.inventory.rest.requests=DEBUG`
* adding a way to disable the agent subsystem by Java system property 
`-Dhawkular.agent.enabled=false`

Agent should not be disabled by default, it just makes things faster if you want to skip the discovery process and test something unrelated. I think, this could be handy also for the integration tests, where we are creating some entities in the inventory and then checking their presence. If the agent is disabled during this process it should remove a lot of side-effects caused by the agent feeding the resources/types/etc. during this tests.

(both params should be passed to the `standalone.sh` not to the mvn)

### DIFF
--- a/dist/src/main/resources/wildfly/patches/standalone.xsl
+++ b/dist/src/main/resources/wildfly/patches/standalone.xsl
@@ -95,6 +95,11 @@
         <xsl:attribute name="name"><xsl:text disable-output-escaping="yes">${hawkular.log.inventory:INFO}</xsl:text></xsl:attribute>
       </level>
     </logger>
+    <logger category="org.hawkular.inventory.rest.requests">
+      <level>
+        <xsl:attribute name="name"><xsl:text disable-output-escaping="yes">${hawkular.log.inventory.rest.requests:INFO}</xsl:text></xsl:attribute>
+      </level>
+    </logger>
     <logger category="org.hawkular.metrics">
       <level>
         <xsl:attribute name="name"><xsl:text disable-output-escaping="yes">${hawkular.log.metrics:INFO}</xsl:text></xsl:attribute>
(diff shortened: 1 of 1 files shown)


dist/src/main/resources/wildfly/patches/standalone.xsl
```diff
@@ -349,7 +354,7 @@
                  numAvailSchedulerThreads="3">
         <xsl:attribute name="enabled">
           <xsl:choose>
-            <xsl:when test="$kettle.build.type='dev'">true</xsl:when>
+            <xsl:when test="$kettle.build.type='dev'">${hawkular.agent.enabled:true}</xsl:when>
```
metlos: shouldn't this be also wrapped in `<xsl:text disable-output-escaping="yes">` as it is done for logger levels?

Jiri-Kremser: no need, http://www.w3schools.com/xsl/el_text.asp

yes means no escaping of special chars `<` -> `&lt;`.
Here the values that makes sense for this parameter is `true` and `false` and if some possible attacker wants to inject some tricky xml snippet by some kind of xml injection it's imho better to escape the 'picitka' (<> :])

metlos: @ppalaga what was the reason for adding the `<xsl:text disable-output-escaping="yes">` for the log levels?

ppalaga: @metlos IIRC, it is so that the dollar sign does not have its XSL meaning and is rendered as literal

metlos: which would make it required in your case, @Jiri-Kremser, I think?

ppalaga: Escaping of dollar signs in XSL is tricky. I have run the build like this and it produces what it was supposed to produce. Let's merge like this and revisit later if necessary.

Jiri-Kremser: thanks for the merge
I think it's because the `$`, `{` and `}` aren't named entities in XML (but <, >, &, etc. are)

PR #390, opened 2015-08-04
Inventory version++

Jiri-Kremser:
+ reflecting the changes in Maven module names in the inventory


ppalaga:
I have built and run with this change and everything seems to work

PR #393, opened 2015-08-05
Import WildFly BoM

ppalaga:


ppalaga:
@pilhuhn could you please review and merge?

PR #397, opened 2015-08-11
HAWKULAR-529 Log the stack trace in case of ping IOException 

ppalaga:


ppalaga:
@jpkrohling could you please review?

jpkrohling:
Looks good to me. I'd track the other "feature requests" in other JIRAs, such as "different states than UP/DOWN" and "On the first failure, the pinger could record the problem as ERROR on the logs, so that an admin could at least get a hint on what's going on".

PR #402, opened 2015-08-14
HAWKULAR-430 add-user Wildfly broken due to missing 'domain' directory

ppalaga:


ppalaga:
@jpkrohling this brings back just the `*.properties` files under `*/domain/**` back, so that `add-user.sh` & Co. do not complain. 

All other files under `*/domain/**`are still filtered out so that we do not awake an impression that domain setups are supposed to work out of the box.

OK like that?

jpkrohling:
That would solve this specific problem. But I'm afraid that other tools might complain about other files. Could you check on JBoss Tools if it detects Hawkular as a regular Wildfly app server? 

For IntelliJ, it works fine the way it is, ie, I can add Hawkular as a local and/or remote server and I can both start and debug it, as long as the domain mode is not used.

ppalaga:
Installing JBoss Tools...

ppalaga:
@jpkrohling it is up to you to choose standalone or domain when creating the server profile http://snag.gy/tO13R.jpg - i.e. not having domain does not cause anything fatal in JBoss Tools.

jpkrohling:
Looks good then.

PR #403, opened 2015-08-14
HAWKULAR-553 Additional condition to skip adding invalid URL

vrockai:
I don't have Safari to test this and I don't know why it submits an invalid form, but this should fix the issue (although may not be the ideal way to do so).


mtho11:
Merged #403

PR #406, opened 2015-08-18
HAWKULAR-543 : Deployments "Select All" doesn't work

ammendonca:
Fixes included:
- Toggling checkbox selects the row (and visually shows as selected);
- Clicking the row also selects it;
- Switch info bar to "N elements selected" when there's a selection;
- Select All toggles between selecting all and none;
- Enable/Disable multi operation button if there is/isn't a selection;
- Multi operation works on selected resources instead of all;


mtho11:
Merged #406

PR #411, opened 2015-08-24
HAWKULAR-503 - Red Hat Access Integration

jpkrohling:
I'm sending this as a PR to get a feedback both on the code as well as the feature. There are a couple of issues that I've found during the testing, which I don't think are blockers. I'll try to get them fixed today, but they might get done only on the next milestone.

Known issues:

- End user needs to click twice at the "Red Hat Access" toolbar option for the dropdown to open for the first time
- Once logged in, when loading a new option on the toolbar menu, the end user has to wait for the login procedure to happen in background. 
- Upload doesn't work when opening a case. Workaround: upload the files once the case is already open.


jpkrohling:
@pilhuhn , @theute , could you please test this branch? 
@mtho11 , could you please review the code for the PR?

theute:
Beside the known issues, the hawkular header is hidden when accessing the access portal (can be seen by scrolling up) and I have up to 3 scrolling bars http://snag.gy/r1YRl.jpg (vs 2 in JON3 http://snag.gy/kSc47.jpg).

Chrome Version 44.0.2403.125 (64-bit)

jpkrohling:
This last two commits to this PR should fix the following issues:

* Need to click twice to open the Red Hat Access dropdown
* Scroll issue on the "main" window. The iframe size is now calculated to be 80% of the visible browser screen, as proposed by @pilhuhn

console/src/main/scripts/plugins/accessPortal/html/open_case.html
```diff
@@ -0,0 +1,3 @@
+<iframe id="openCaseWindow" width="100%" frameborder="0" src="/support.html#/case/new" scrolling="auto"
+        data-ng-controller="AccessPortal.AccessPortalMainController as accessPortalMain">
```
mtho11: Small nit. We have been removing the data-* to all the ng-* directives. This can be cleaned up later though.

jpkrohling:
Commits squashed. If Travis approves (which it did on the separate commits), this is ready to be merged.

mtho11:
#411

PR #413, opened 2015-08-24
Fix Typescript tslint error: 

mtho11:
[gulp-tslint] error accessPortal/ts/accessPortalPlugin.ts[44, 26]: missing radix parameter


mtho11:
I'll wait to merge this until after the build is fixed.

mtho11:
Yes,  this is the original error:
[ERROR] plugins/accessPortal/ts/accessPortalPlugin.ts(43,35): error TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.

Looks like I pasted the radix error above (incorrectly) instead of the original. The missing radix looks to be a bug in the typescript definition files because the radix is optional:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

jpkrohling:
I sent a new PR for fixing this issue (see #419) . I think this PR could be then closed.

mtho11:
Close due to fix in #419

PR #417, opened 2015-08-26
Correct some paths which dont end with ~~ but with ~/ instead

ammendonca:
Also update agent to latest version


ppalaga:
Replaced by #420

PR #422, opened 2015-08-26
HAWKULAR-579 : Split Alert settings into Response Time and Availability

ammendonca:


ppalaga:
Yes, this gets rid of the broken link, but do you find this nice enough so that it can be considered as a permanent solution? 
@mtho11 wdyt? this is the present PR:
image

ppalaga:
I merged, but the question whether this solution should be permanent is still open.

ammendonca:
@ppalaga you are right, it's not good enough, but it's the quick way. @cardosogabriel should help on finding a better solution (likely will consist on a new modal with both, separated by tabs)

PR #429, opened 2015-08-31
HAWKULAR-556 - Attach Hawkular's JDR to case.

jpkrohling:
@lzoubek , would you please review this one? It's mostly based on your code, with minimal changes.

@theute / @pilhuhn , although I tested it locally, it would be interesting if you could test it as well, to check in other browsers/OSes. I tested in Firefox and Chrome for Fedora 22.

### DIFF
--- a/console/src/main/scripts/plugins/accessPortal/vendor/js/custom-cases.js
+++ b/console/src/main/scripts/plugins/accessPortal/vendor/js/custom-cases.js
@@ -1,10 +1,12 @@
+var attachmentsRequest = 'hawkular-redhat-access-integration-backend/attachments';
+
 angular.module('RedhatAccess.JON', ['RedhatAccess.cases'])
 .controller('customCase', ['$scope', 'securityService', 'NEW_DEFAULTS', '$location', '$http', function($scope, securityService, NEW_DEFAULTS , $location, $http) {
   NEW_DEFAULTS.product = "Red Hat JBoss Operations Network";
   NEW_DEFAULTS.version = "3.3.0";
 
   var params = $location.search();
-  
+
   // handle support case for managed resource
   if ($location.path().indexOf('/resource-case') >= 0) {
     NEW_DEFAULTS.product = params.product;
@@ -16,3 +18,19 @@ angular.module('RedhatAccess.JON', ['RedhatAccess.cases'])
      securityService.validateLogin(true);
    };
 }]);
+
+angular.module('RedhatAccess.cases')
+  .controller('BackEndAttachmentsCtrl', ['$scope', 'TreeViewSelectorData', 'AttachmentsService',
+    function ($scope, TreeViewSelectorData, AttachmentsService) {
+      $scope.name = 'Attachments';
+      $scope.attachmentTree = [];
+      TreeViewSelectorData.getTree(attachmentsRequest).then(
+        function (tree) {
+          $scope.attachmentTree = tree;
+          AttachmentsService.updateBackEndAttachments(tree);
+        },
+        function () {
+          console.log('Unable to get tree data');
+        });
+    }
+  ]);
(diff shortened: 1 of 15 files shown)


modules/redhat-access-integration-backend/src/main/java/org/hawkular/accesportal/backend/control/RedHatAccessConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+    private String userAgent;
+    private String url;
+    private String proxyUser;
+    private String proxyPassword;
+    private URL proxyUrl;
+    private int proxyPort;
+
+    public RedHatAccessConfiguration() throws MalformedURLException {
+        this.brokered = true;
+        this.userAgent = "redhat-access-plugin-jon-1.0.4";
```
lzoubek: Could you please set userAgent to (ie. redhat-access-plugin-hawkular-1.0.4). UserAgent is used by RHA portal to generate analytics and should be in form "redhat-access-plugin-$product-$javaLibVersion"

jpkrohling: Sure, will do. I wasn't sure if I should use JON's, as there might be some restrictions on the server side related to unknown UAs, but will do it.

modules/redhat-access-integration-backend/src/main/java/org/hawkular/accesportal/backend/boundary/AttachmentEndpoint.java
```diff
@@ -0,0 +1,91 @@
...
+                selectedFileForUpload.getPassword(),
+                configuration.getUrl(),
+                configuration.getProxyUser(),
+                configuration.getProxyPassword(),
+                configuration.getProxyUrl(),
+                configuration.getProxyPort(),
+                configuration.getUserAgent(),
+                configuration.isDevel()
+        );
+
```
lzoubek: in JON we test API connection/authentication first via ```api.getProblems().diagnoseStr("test");``` - you don't need to create JDR report and then eventually fail to upload it.

jpkrohling: Alright, I've seen that code but wasn't sure it had a purpose. It seems it have :)

jpkrohling:
Rebased, made the changes based on the review and squashed again. 
@lzoubek , could you please check it?

lzoubek:
looks good to me

console/src/main/scripts/plugins/accessPortal/vendor/js/redhat_access_angular_ui.js
```diff
@@ -3522,7 +3522,7 @@ angular.module('RedhatAccess.cases').controller('EditGroup', [
         $scope.accountNumber = null;
         $scope.isUsersPrestine = true;
```
mtho11: Ultimately, lets get rid of this $scope syntax and use 'ControllerAs' syntax instead.

console/src/main/scripts/plugins/accessPortal/vendor/js/custom-cases.js
```diff
@@ -1,10 +1,12 @@
+var attachmentsRequest = 'hawkular-redhat-access-integration-backend/attachments';
+
```
mtho11: A javascript project and not typescript? All of the other plugins are typescript.

mtho11:
Merged #435

console/src/main/scripts/plugins/accessPortal/vendor/js/custom-cases.js
jpkrohling: That's because those JavaScripts are not coming from us *or* are very temporary, so, it's a sign that we should throw it away once a proper solution is offered by the Red Hat Access folks. This code is 90% the one that exists for RHQ.

console/src/main/scripts/plugins/accessPortal/vendor/js/redhat_access_angular_ui.js
jpkrohling: This JS is not ours, it's a library from Red Hat Access, which is not bowerized.

console/src/main/scripts/plugins/accessPortal/vendor/js/custom-cases.js
mtho11: Ah ok.

It would be nice if they could package as bower package though.

PR #432, opened 2015-09-01
HAWKULAR-581 needed to secure the websocket comm

jmazzitelli:


mtho11:
Doesn't look like deploy operations or deploy add are enabled to use WS security.

mtho11:
Breaking change.

PR #439, opened 2015-09-08
trying to get the new command-gateway stuff into hawkular

jmazzitelli:

PR #440, opened 2015-09-08
[HWKIVENT-133] Make the URLs sorted according to custom UI-specific rules.

metlos:
The reason why this is completely done in the UI is that a) this type of sorting is specific to the UI - no other client will most probably benefit from it and b) the UI merely defines the data to be used for sorting and then instructs the inventory to sort using it.


mtho11:
LGTM.

mtho11:
Merged #439

mtho11:
Totally agree that is a very UI centric thingy.

PR #441, opened 2015-09-08
Upgrade agent and command-gateway to the last revisions

ppalaga:


ppalaga:
@jmazzitelli after I had no success with running the cmdgw itests here, I moved them to agent, see https://github.com/hawkular/hawkular-agent/pull/45

So now the present PR just upgrades agent and cmdgw.

PR #461, opened 2015-09-15
ES6 'let' is the new 'var'

mtho11:
For reasons detailed here: https://hacks.mozilla.org/2015/07/es6-in-depth-let-and-const/


mtho11:
Merged #461

PR #471, opened 2015-09-16
Remove unused methods for actions manipulation

lucasponce:
Minor refactor of actions.,
With this alertManagers.ts is the single point from UI to manipulate Alerts API, then future upgrades will be more easy to maintain.


mtho11:
Merged #471

PR #472, opened 2015-09-17
HAWKULAR-610 File chooser for add-deployment only accepts .zip by def…

mtho11:
…ault on some browsers.


mtho11:
@pilhuhn Looks like the 'application/java-archive' doesn't work either.

pilhuhn:
We should probably just accept everything in the file chooser.

mtho11:
For now, I agree.

mtho11:
I had to move this to #483 because of rebase issues on that branch.

PR #478, opened 2015-09-18
HAWKULAR-465 Add Web Tab alert dialog for Sessions

lucasponce:
This PR is focus on the Web Session metrics, as those are present in the page.
The original JIRA covers also Container Threads and Requests which are not shown.
So, I decided to support only alerts with visible metrics.
Please @mtho11 and @jshaughn take a look at this work and merge if you think is correct.


mtho11:
LGTM.
@ammendonca can you answer Lucas's question regarding Container Threads and Requests which are not shown. I can't remember what the reason was.

ammendonca:
IIRC those metrics were not available.. not sure it is still the case (or if I missed them)

PR #479, opened 2015-09-19
HAWKULAR-614 Web socket closing because of no auth. Also resourceId c…

mtho11:
…hanged to resourcePath. Adding additional typesafety around Resource List.


mtho11:
Wow, finally. After 4 builds!

PR #485, opened 2015-09-21
HAWKULAR-507 : Close dropdown after selecting deployment action

ammendonca:
Also fix showing selected rows in blue


mtho11:
Merged #485

PR #489, opened 2015-09-22
HWKALERTS-77 Add resourcePath into Trigger.context

lucasponce:
Triggers and Conditions contexts are designed to be used at client level, so any additional field can be added easily there, as it doesn't affect into the alerts inference process.
please, @mtho11 take a look to see if it suits your requeriments for Alert Center API.


lucasponce:
I don't understand why travis is failing, perhaps the src dependencies in pom.xml ?
I could run a "mvn clean install" several times locally.

mtho11:
Sometimes it takes a few times to pass.

mtho11:
Merged #484 

Thanks @lucasponce !

PR #491, opened 2015-09-22
Fix Deployment file name cleaning

ammendonca:
Not all browsers append the "C:\fakepath\" and, for those who appended
if going back and forth, filename was getting cut by 12 chars every time


mtho11:
Nice catch @ammendonca !

PR #503, opened 2015-09-30
HAWKULAR-596 : Improved counter/rate data at GC

ammendonca:
Populating missing data (for start and in the middle) with 0 values so
that time range and generated buckets are good.

Also upgraded to Hawkular Metrics 0.6.0.Final

### DIFF
--- a/console/src/main/scripts/plugins/metrics/ts/app-details/appServerJvmDetails.ts
+++ b/console/src/main/scripts/plugins/metrics/ts/app-details/appServerJvmDetails.ts
@@ -143,21 +143,57 @@ module HawkularMetrics {
     }
 
     private formatCounterChartOutput(response, buckets = 60):IChartDataPoint[] {
-      let result = response;
+      if(response.length < 2) {
+        return [];
+      }
+
+      // get the timestamp interval from the first two samples
+      let tsStep = response[1].timestamp - response[0].timestamp;
+
+      // sometimes there are gaps in data, which needs to be filled with empty values so the buckets get similar time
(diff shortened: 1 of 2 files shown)


console/src/main/scripts/plugins/metrics/ts/app-details/appServerJvmDetails.ts
```diff
@@ -143,21 +143,57 @@ module HawkularMetrics {
     }
 
     private formatCounterChartOutput(response, buckets = 60):IChartDataPoint[] {
```
mtho11: Can this go in the MetricsService as it should be reusable for all counters
https://github.com/hawkular/hawkular/blob/ee54b476d14c039f5bb5df908ecab44ce547ae31/console/src/main/scripts/plugins/metrics/ts/services/metricsService.ts#L82-82
Also, can you add a retrieveCounterMetrics method.
This can be in a separate commit.

ammendonca: This is something supposed to be (very) temporary, because once metrics have support for buckets in counters most of it will go away, the remaining will go with the ability of charts to handle metrics unmodified [bucket] data, as per https://issues.jboss.org/browse/HAWKULAR-643.

Still, this can go in the MetricsService, don't think there's anything specific to it that'd block it. Moving.

mtho11:
@ammendonca can you rebase again?

ammendonca:
New PR https://github.com/hawkular/hawkular/pull/504. Closing this one.

PR #512, opened 2015-10-01
Review HAWKULAR-67 implementation

cardosogabriel:


cardosogabriel:
Well done Juca! I comment inline:

- Dropdown switch: please follow the design specified in hawkular-67 + Manage Organizations link
- Spinner loading: use the same approach as Application Servers, for example. The title of the page should appear also only after the spinner.
- Create organization: the feedback (toastr) should be success and text: “Organization successfully created.”
- Organizations table: since we have a big space, we could list the members as hyperlinks, like:
	You (admin), Juraci, Thomas and 1 more 
	Clicking on one name would go to edit user. Clicking on 1 more would go to the users list screen.
- Organization table: tooltip Delete is not showing on the trash icon. Need fix.
- When removing an organization, we should have the confirmation dialog like in the urls list. Also, the feedback (toastr) should be success type (not info), and text: “Organization successfully deleted.”
- Organization not found: no need for feedback tostr on this page.

jpkrohling:
Are the changes to the metrics less file intentional? Wouldn't they influence the look and feel of metrics? Apart from that, I'm ok with this PR.

console/src/main/scripts/plugins/metrics/less/metrics.less
```diff
@@ -235,7 +251,7 @@ a:hover {
 
   .hk-graph-container  {
     margin: 0 auto;
-    //width: 750px;
+    width: 750px;
```
mtho11: Wouldn't this return us to fixed width graphs? And not the semi-responsive that we have now?

cardosogabriel:
Yes, they are intentional :)

mtho11:
@cardosogabriel looks like this needs to be rebased to master

mtho11:
Merged #512

PR #514, opened 2015-10-01
Remove subdirs of $HOME/.m2/repository older than 10 days in Travis

ppalaga:


ppalaga:
@pilhuhn there is [Don't Merge Yet], the problem is still not solved.

PR #519, opened 2015-10-05
HAWKULAR-59 - List pending invitations

jpkrohling:


console/src/main/scripts/plugins/accounts/ts/accountsGlobals.ts
```diff
@@ -63,4 +93,28 @@ module HawkularAccounts {
...
+    role:HawkularAccounts.IRole;
+    acceptedAt:Date;
+
+    constructor(email:string, role:HawkularAccounts.IRole) {
+      this.email = email;
+      this.role = role;
+    }
+  }
+
+  export class Role implements IRole {
```
jpkrohling: I need a concrete class, so that I need to create this object "on demand". It duplicates most of the code from IRole (same for the above Invitation vs. IInvitation), but I think this is a necessary evil, right?

mtho11: Shouldn't be any need to duplicate. Interfaces can extend other interfaces and classes can even have 'abstract' classes now.

mtho11:
Merged #519

PR #521, opened 2015-10-05
HAWKULAR-666 - configure agent for platform metrics

jmazzitelli:


dist/src/main/resources/wildfly/patches/standalone.xsl
```diff
@@ -972,6 +972,13 @@
 
         </managed-servers>
 
+        <platform enabled="true">
+          <file-stores   enabled="true" interval="5" timeUnits="minutes" />
+          <memory        enabled="true" interval="1" timeUnits="minutes" />
+          <processors    enabled="true" interval="1" timeUnits="minutes" />
+          <power-sources enabled="true" interval="5" timeUnits="minutes" />
```
pilhuhn: Are you sure we should enable that by default?

jmazzitelli: No, I am not sure :) We can easy change these defaults to whatever we deem the most appropriate. I just had them all enabled for now as I was testing my implementation

PR #523, opened 2015-10-06
Bump alerts to 0.5.x using srcdeps

lucasponce:
With the upgrade of bus v0.7.0 alerts needed to be aligned and use same API.
I've upgraded into 0.5.x and using srcdeps to point specific commit as we have not finished the 0.5.0 version yet.


pilhuhn:
Thanks @lucasponce

PR #524, opened 2015-10-06
HAWKULAR-60 - Ability to change an user's role in an organization

jpkrohling:


jpkrohling:
@mtho11 , once travis approves, could you please review/merge this one?

console/src/main/scripts/plugins/accounts/html/organization-memberships.html
```diff
@@ -51,7 +51,34 @@ <h1 class="text-center">{{controller.organization.name}} Members</h1>
...
+                {{membership.role.name}}
+              </span>
+
+              <span ng-show="controller.membershipsToUpdate[membership.id] === PersistenceState.PERSISTING">
+                <i class="fa fa-spinner fa-spin"></i>
+              </span>
+              <span ng-show="controller.membershipsToUpdate[membership.id] === PersistenceState.SUCCESS">
+                <i class="fa fa-check"></i>
+              </span>
+              <span ng-show="controller.membershipsToUpdate[membership.id] === PersistenceState.ERROR">
```
jpkrohling: It's because of those lines that we need the PersistenceState on the $rootScope.

console/src/main/scripts/plugins/accounts/ts/accountsGlobals.ts
```diff
@@ -60,9 +60,13 @@ module HawkularAccounts {
   }
 
   export interface IOrganizationMembership {
+    id: string;
     organization: IOrganization;
     member: IPersona;
     role: IRole;
+    $update(options:{},
+            success?:(success:ISuccessPayload) => void,
```
mtho11: ISuccessPayload and IErrorPayload might be a bit too generic. Perhaps something like IOrgMembershipSuccess or something like that...

console/src/main/scripts/plugins/accounts/ts/organizationMembership.ts
```diff
@@ -138,6 +151,19 @@ module HawkularAccounts {
       );
     }
 
+    public loadPermissionToTransferOrganization(organizationId:string):void {
+      let operationName:string = 'organization-transfer';
```
mtho11: const ?

jpkrohling: +1

console/src/main/scripts/plugins/accounts/ts/accountsGlobals.ts
jpkrohling: For the current implementation, it's not needed, as we are not reading results from the server. I think other operations indeed have a more specialized type, but I don't think it's needed for this specific operation.

mtho11:
Merged #524

PR #532, opened 2015-10-07
HAWKULAR-538 Design adjustments

cardosogabriel:
Mike, I saw some things that need fix:
- when selecting an item in the alerts list, the background is not getting dark blue. The class "hk-selected" needs to be applied in the <tr> element.
- In the alert details, after acknowledging an alert, the button is still active. Should be disabled as in the list.
- Popovers don't work on this screen.


mtho11:
Merged #532

PR #534, opened 2015-10-08
HAWKULAR-661 - UI Adjustments

jpkrohling:


jpkrohling:
@cardosogabriel , pretty much all of the items are implemented, except this:

" Organizations table: since we have a big space, we could list the members as hyperlinks, like: You (admin), Juraci, Thomas and 1 more Clicking on one name would go to edit user. Clicking on 1 more would go to the users list screen."

I'm not sure what you mean by that. 

Also, the first item (dropdown switch) has some other issues now, but I think it's mostly CSS-related. I think you'd be more suitable to fixing those :) In any case, I added some CSS classes there already, to make it easier to fix it.

mtho11:
Merge #534

PR #542, opened 2015-10-09
HAWKULAR-538 Additional changes to firstUpper filter and truncate in …

mtho11:
…middle filter along with some state change bug fixes with resolutions to ack and resolved. Specifically, added ability to truncate at the end and in middle as specified by UxD.


mtho11:
Merged #542

PR #546, opened 2015-10-09
HAWKULAR-680 and organizations improvements

cardosogabriel:
Refined:
- Account switch 
- Accept invitation screen
- Feedback messages

Pending, for Juraci:
- When changing user role: show feedback success via toastr "User role successfully changed." Thus, you can remove the check icon after transferring.
- Loading should be the patternfly style: https://www.patternfly.org/widgets/#spinner
- Select should be patternfly style: https://www.patternfly.org/widgets/#bootstrap-select
- The feedback message in the modal when transferring ownership should contain the line below: “This action can’t be undone.” See modal delete url for reference


mtho11:
Merged #546

PR #548, opened 2015-10-09
Reduce the size of the angular generated DOM and increase runtime per…

mtho11:
…formance by turning off angular debug info.

Angular spits out all kinds of debug info that is not wanted in a production app. This info can easliy be turned back on at runtime by typing at the js console: angular.reloadWithDebugInfo().


mtho11:
Merged #548

PR #555, opened 2015-10-13
Add partial sorting on AlertsCenter

lucasponce:
This PR implements sorting from server side.
There are pending tasks to be done, for example, it is needed to combine pagination + sorting.
This is not yet implemented but instead to send a full large commit I'm going to send small ones.
On this way, it is easy to cut for the MS having things on a stable way.
@mtho11 can you take a look, please ?
Thx.


mtho11:
LGTM.

PR #556, opened 2015-10-13
HAWKULAR-557 Broken on Windows

jshaughn:
The use of ${jboss.server.data.dir} in a URI inside of standalone.xml is
a problem. The windows path backslashes are not valid.  Instead, use the
new org.hawkular.data.dir, defaulting back to the jboss property.  This
allows windows users to set the new property on the command line to
bypass the issue (or to use a directory not under the jboss server,
if desired)


tsegismont:
We live in a sad, sad world, where people spend time developing bots 
like this one :)
https://twitter.com/brokencodebot/status/654001268709531648

theute:
lovely Twitter account :)

PR #560, opened 2015-10-14
Move alerts actions into a dropdown

lucasponce:
Last minor to have updated the same style as presented by UI for "View Alerts" umbrella.
There are pending things not yet implemented as we need support from backend:
- Sorting by trigger context is not yet supported in Alerts.
- Comments are only supported on transition state but not as generic ones per alert, this need also support on the backend.

@mtho11 please, if you can take a look and merge, with this the AlertCenter has a good progress in this area.

I'm going to move into Hawkular Alerts to add these pendings services, back into UI when ready.


mtho11:
Merged #560

PR #544, opened 2015-10-09
HWKINVENT-149 | inventory itests moved to inventory

pavolloffay:

PR #562, opened 2015-10-15
[HWKINVENT-149] Import inventory itests from inventory module

metlos:
This is an updated version of PR #544.
All it does is that it updates the version of inventory used in dist to 0.6.0.Final. With this, the end-to-end tests defined in this repo pass fine and also the correct version of inventory integration tests is included and run against the full dist.


ppalaga:
Lgtm, thanks guys!

PR #574, opened 2015-10-16
Update the hawkular-ui-services for platform tab (experimental).

mtho11:


mtho11:
Merged #574 

@jshaughn the right ui-services should now be available

PR #576, opened 2015-10-19
HAWKULAR-676 - Topology view

Jiri-Kremser:
It's enabled only under experimental mode. To enable experimental mode, click on the hawkular brand (top left).


mtho11:
Should this be a standalone plugin? Its just a tab right? Its much easier to do integrations for the future if its codebase is part of Hawkular. And we have explicit control over the css(less) styles loading overriding one another.

Also, shouldn't the directives be part of Hawkular-charts?

PR #586, opened 2015-10-20
Move calls to (bucketed) metrics all to MetricsService

ammendonca:
Non-bucketed calls are still not moved as it needs further changes. Also
moved all calls to formatBucketedChartOutput to use method from service.


mtho11:
Merged #586

PR #593, opened 2015-10-20
Hawkular 654

jshaughn:
The triggers list portion of 654.
- Not yet forwarding to a details page.

### DIFF
--- a/console/src/main/scripts/plugins/metrics/html/alerts-center-triggers.html
+++ b/console/src/main/scripts/plugins/metrics/html/alerts-center-triggers.html
@@ -34,13 +34,12 @@
       <div class="hk-table-container">
         <div class="hk-actions-table">
           <div class="dropdown">
-            <button class="btn btn-default dropdown-toggle" type="button" id="dropdownMenu1" data-toggle="dropdown">
-              Change State
-              <span class="caret"></span>
-            </button>
-            <ul class="dropdown-menu" role="menu" aria-labelledby="dropdownMenu1">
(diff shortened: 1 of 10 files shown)


console/src/main/scripts/plugins/metrics/html/alerts-center-triggers.html
```diff
@@ -0,0 +1,93 @@
...
+          <tr ng-repeat="trigger in act.triggersList | filter:act.search" ng-class="{'hk-selected': act.selected}"
+              ng-click="act.selectItem(trigger)">
+            <td><input type="checkbox" ng-checked="trigger.selected"/></td>
+            <td>{{trigger.name}}</td>
+            <td>{{trigger.description}}</td>
+            <td>{{trigger.context.resourceName|truncate: 30}}</td>
+            <td>{{trigger.severity|firstUpper}}</td>
+            <td>{{trigger.enabled ? "True" : "False"}}</td>
+            <td>{{trigger.actions.email.length}} <i class="fa fa-envelope"></i></td>
+            <td><a class="btn btn-link"><i class="fa fa-chevron-circle-right fa-lg"
```
lucasponce: The PR looks good to me.
The only comment I have is that as this is going to be iterative, if for this cut we are not going to add the detail page I should hide/comment the icon to not present a link that has not defined the alert page.
So in that case, the MS6 would have the trigger list but that's all (and it's ok) we can complete it later.
But it's a minor comment, no blocker, I'm +1 to merge it as it is and refine this change later of this PR.

mtho11:
LGTM for first iteration. I agree with @lucasponce  comment above about detail link.

mtho11:
Merged #592

PR #606, opened 2015-10-22
HAWKULAR-745 : Update angular-wizard so wz-disabled is supported

ammendonca:
Also remove misleading patternfly resolution.


mtho11:
Merged #606

PR #617, opened 2015-10-28
HAWKULAR-758 - Added cassandra and driver as JBoss Modules

jpkrohling:


jpkrohling:
This does *not* affect other modules, as they will still use whatever they have on their WARs. If they chose to use the module as a dependency, Accounts can be used as reference.

dist/assembly.xml
```diff
@@ -161,6 +161,30 @@
       <useTransitiveFiltering>false</useTransitiveFiltering>
     </dependencySet>
 
+    <!-- Cassandra module -->
+    <dependencySet>
+      <outputDirectory>${hawkular.dist.zip.root.dir}/modules/org/apache/cassandra/main</outputDirectory>
+      <includes>
+        <include>org.apache.cassandra:cassandra-all:jar</include>
```
pilhuhn: What is that? Basically the cassandra.ear as a module now?

jpkrohling: That contains Cassandra, but it's not started as embedded. The embedded EAR still bundles its own Cassandra (which should be the same version, as it's defined on the parent pom). Basically, it's there as a module so that components can use its API. 

IIRC, we don't use the individual modules from Cassandra, only the -all artifact, so, that's the one that's added as a module.

jpkrohling: By the way: I see no problem in changing the EAR to consume this Cassandra from the module, but I'd do it in phases. First, this change that should not affect other modules, then change the EAR, which should affect other modules.

tsegismont: I don't understand why a component other than inventory needs cassandra-all. A module for the driver, why not. But cassandra-all, it's another story.

jpkrohling: Indeed, for some reason, I thought I needed access to a class or two inside this jar, but I was wrong. I'm going to remove this and update the PR.

jpkrohling:
As discussed with @tsegismont , the module will be added to the nest. This PR will be updated to contain the SRC-revision for the nest revision, once it's reviewed/merged there.

jpkrohling:
The PR has a reference to the SNAPSHOT, but will be updated to the actual version once the PR on Bus is merged. The contents of this PR were updated only for review purposes.

dist/assembly.xml
```diff
@@ -141,6 +141,7 @@
       <includes>
         <include>org.hawkular.agent:hawkular-monitor:zip:module</include>
         <include>org.hawkular.bus:hawkular-bus-ra-wf-module:zip:module</include>
+        <include>org.hawkular.bus:cassandra-driver-wf-module:zip:module</include>
```
tsegismont: If it's already in the nest zip, why do we need to specify it here?

jpkrohling: It's in the hawkular-nest-distro, which doesn't seem to be consumed by Hawkular. Hawkular consumes the individual pieces of nest.

tsegismont: Le 29/10/2015 10:12, Juraci Paixão Kröhling a écrit :
> It's in the hawkular-nest-distro, which doesn't seem to be consumed by
> Hawkular. Hawkular consumes the individual pieces of nest.

That's a surprise to me. Maybe @mazz knows why?

mazz: I'm not on this project.

tsegismont: @mazz sorry, I meant @jmazzitelli , our "mazz" :)

mazz: No problem.

mazz:
Take me off this DL please.

mazz:
Nevermind, I did it ..

PR #620, opened 2015-10-29
use the new agent module dist

jmazzitelli:


dist/pom.xml
```diff
@@ -42,8 +42,7 @@
   <dependencies>
     <dependency>
       <groupId>org.hawkular.agent</groupId>
-      <artifactId>hawkular-monitor</artifactId>
-      <classifier>module</classifier>
+      <artifactId>hawkular-monitor-wf-extension</artifactId>
```
jpkrohling: <exclusions> here :)

PR #623, opened 2015-10-31
Move charts back to 0.5.19 instead of snapshot

mtho11:
This is because we are doing some breaking changes on master now.


mtho11:
Merged #623

PR #628, opened 2015-11-03
HAWKULAR-767 - Added dozip profile to .travis.yml

jpkrohling:


Jiri-Kremser:
It shouldn't be necessary to use the dozip profile. It creates the zip and tar by default.

-Pdev .... create only the dir
-Pdozip .... create the dir, zip and tar
nothing  .... zip and tar

jpkrohling:
Indeed. I just tried locally, and it seems to work fine without this option. I also tried to start Hawkular from the ZIP it generated from a simple "mvn clean install", and it worked fine.

PR #630, opened 2015-11-04
HAWKULAR-767 - Perform build and deploy in one step

jpkrohling:


jpkrohling:
@ppalaga , @stefannegrea , @Jiri-Kremser , not sure if this was done/tested like this before, but it should solve the problem of processing standalone.xml twice (once during the "script" part, and once again at the "after_success" part). The side effect is that it would fail the build if there's a failure during the deployment, but I think this is actually a good thing, as we would then be sure that the "latest" would be deployed. What do you think ?

jpkrohling:
https://issues.jboss.org/browse/HAWKULAR-780 - For a more longer term fix. This should fix the issues that QE is having with living on the edge.

PR #633, opened 2015-11-05
Bump Hawkular-charts to 0.5.21.

mtho11:


mtho11:
Merged #633 
That took 3 tries to pass integration tests.

PR #637, opened 2015-11-06
HAWKULAR-518 : Fix tooltip cropped on "Update chart"

ammendonca:
Also fixed it in other places where this was/might happen.
Took the chance to also do some general improvements and respect line length in affected files.


mtho11:
@ammendonca can you now rebase and resubmit, now that we should have things working...

mtho11:
Merged #637

PR #638, opened 2015-11-09
Upgrading the inventory version to 0.8.0.Final and the agent version …

Jiri-Kremser:
…to 0.12.0.Final


Jiri-Kremser:
Ok, this fixes the itest http://pastebin.com/DhUMF8Nf but I guess it's not a good idea to do that, because then it'd fail in the UI. Namely here: http://git.io/v8apR

@metlos any ideas why the relative path doesn't work? I recall, I fixed the itests for the inventory to make them pass without paying the attention to the possible issue (here http://git.io/v8ajB), I should have pinged you earlier, something looks odd here. iirc, the server side was saying something like: "it is ambiguous to guess the path", is it possible that the "feeds -> tenant" move caused this?

Hot-fix would be to apply the patch + changing it in the urlList.ts in the same way.

mtho11:
Merged #638

PR #639, opened 2015-11-09
Hawkular 500

karelhala:
Add uglify and ng-annotate to plugins, enable minification in concat file. Add Injections in PersonaInterceptorService and to d3setup. Change controller to link in HkDataPagination and HkPagination.


mtho11:
Merged #639

PR #641, opened 2015-11-10
HAWKULAR-794 Fixing the icons in the topology graph on FireFox

Jiri-Kremser:
* Instead of using the referencing within the svg element, we draw the circle with an fontawesome character over and over again.

* It also fixes the behavior near the viewport edges.


mtho11:
Merged #641

PR #642, opened 2015-11-10
HAWKULAR-500

karelhala:
Injections are now handled with ng-annotate during gulp build, so they can be removed from .ts files. Be careful about variable names , as they should have same name as is registered in module. In this case HawtioNavBuilderProvider was named builder which was not recognized with ng-annotate and threw error.


mtho11:
Merged #642

PR #643, opened 2015-11-11
ui-services tests fails because the http localhost 9876 triplet is no…

Jiri-Kremser:
…t white-listed among the allowed origins (PhantomJS uses the 9876 port)

### DIFF
--- a/dist/src/main/resources/configuration/hawkular-realm-for-dev.json
+++ b/dist/src/main/resources/configuration/hawkular-realm-for-dev.json
@@ -88,7 +88,7 @@
       "bearerOnly" : false,
       "publicClient": false,
       "redirectUris": ["/*", "http://localhost:8080/*", "http://127.0.0.1:8080/*"],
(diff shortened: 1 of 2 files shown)


dist/src/main/resources/configuration/hawkular-realm-for-dev.json
```diff
@@ -88,7 +88,7 @@
       "bearerOnly" : false,
       "publicClient": false,
       "redirectUris": ["/*", "http://localhost:8080/*", "http://127.0.0.1:8080/*"],
-      "webOrigins": ["http://localhost:2772", "http://localhost:8080", "http://127.0.0.1:8080"],
+      "webOrigins": ["http://localhost:2772", "http://localhost:8080", "http://127.0.0.1:8080", "http://localhost:9876"],
```
jpkrohling: Is it for automated tests? If not, then you might also want to add the IP variation to the list: 127.0.0.1:9876.

jpkrohling: Other than that, LGTM

PR #645, opened 2015-11-11
HAWKULAR-788 - Topology view keeps showing the waiting spinner when no resource.

Jiri-Kremser:
Displaying the spinner only during the call of `getData()`. If the inventory is empty, "No Data" is displayed instead of the spinner.

Also making the topology graph visible by default.


mtho11:
Merged #645

PR #646, opened 2015-11-11
HAWKULAR-768 : Show Product Name instead of Resource Type and show Ve…

ammendonca:
…rsion

Fallback to Resource Type in case Product Name is not available.


mtho11:
Merged #646

PR #650, opened 2015-11-11
HAWKULAR-798 Make default time period 12 hours instead of 1 hour. (1 …

mtho11:
…hour didn't really give us anything to drill into).


Jiri-Kremser:
lgtm

PR #654, opened 2015-11-12
Hawkular-571 Alert Query by Resource

jshaughn:
- Replace using multiple triggerId queries with a single resourceId-tag query,
  when fetching alerts in the detail views.
  - We should avoid using IDs for any more than as a unique identifier.
  - We may now be able to make triggerIds UUIDs and not be comprised of
    resourceIds, which would be nicer when we use them in route urls.
  - store alertType on context for easy, consistent, non-hardcoded use.

Also:
- Bump root pom to Alerts 0.6.0.Final
- Do not use context.description for sorting the AC alerts list
  - Instead, use the new native trigger.description support offered by alerts
  - reason 1: context.descroption was not being set on all triggers
  - reason 2: trigger description is editable. The value presented on screen
    may not be the value used for sorting.
  - reason 3: context bloat
- Tweak JVM GC trigger name to start with 'JVM ' for better AC list grouping
- Fix Bug, fix Alert text for Heap alerts
- Fix bug, AC Definitions list State should be Enabled|Disabled, not True|False
- Fix Bug, Datasource triggers did not autodisable (or enable on resolve), make consistent with jvm triggers
- Fix Bug, URL triggers did not turn on autoresolve even though they had autoresolve support in the conditions, now uses autoresolve and also autoenable on resolve
- Fix bug, use ng-span as opposed to span element
- Fix bug, make sure description fits in alert sidebar box
- Fix bug, fix 'Alert Settings' link in URL Response Time tab view
- Fix bug, remove unused query for *all* open alerts
- Fix bug, make sure URL alerts query does not return prematurely


mtho11:
LGTM. Merged #654

PR #658, opened 2015-11-13
Hawkular 806

Jiri-Kremser:
* extending the agent servlet to be able to serve the agent installer. The servlet takes couple of query params and puts their values into property file stored inside the jar so the user then can just download the jar and run 
```bash
java -jar agent-installer.jar
```
* Agent installer UI page with some form fields that are actually mapped into those properties from ^
(the page is by default hidden under the experimental mode)

### DIFF
--- a/dist/assembly.xml
+++ b/dist/assembly.xml
@@ -136,6 +136,19 @@
       <directoryMode>0755</directoryMode>
     </dependencySet>
 
+    <!-- put the agent installer to the same directory as previous artefact from the same reason -->
+    <dependencySet>
+      <outputDirectory>${hawkular.dist.zip.root.dir}/standalone/configuration</outputDirectory>
+      <useProjectArtifact>false</useProjectArtifact>
+      <includes>
+        <include>org.hawkular.agent:hawkular-wildfly-agent-installer:jar</include>
+      </includes>
+      <outputFileNameMapping>agent-installer.${artifact.extension}</outputFileNameMapping>
+      <unpack>false</unpack>
+      <fileMode>0644</fileMode>
+      <directoryMode>0755</directoryMode>
+    </dependencySet>
+
     <!-- TODO Make the unpacking conditional == only for the dev profile -->
     <dependencySet>
       <outputDirectory>${hawkular.dist.zip.root.dir}/modules/org/hawkular/nest/main/deployments/${artifact.artifactId}.war</outputDirectory>
--- a/modules/hawkular-wildfly-agent-download/src/main/java/org/hawkular/component/wildflymonitor/WildFlyAgentServlet.java
+++ b/modules/hawkular-wildfly-agent-download/src/main/java/org/hawkular/component/wildflymonitor/WildFlyAgentServlet.java
@@ -17,13 +17,20 @@
 package org.hawkular.component.wildflymonitor;
 
 import java.io.BufferedInputStream;
+import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Enumeration;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import java.util.zip.ZipOutputStream;
 
 import javax.servlet.ServletException;
 import javax.servlet.annotation.WebServlet;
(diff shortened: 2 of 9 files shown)


modules/hawkular-wildfly-agent-download/src/main/java/org/hawkular/component/wildflymonitor/WildFlyAgentServlet.java
```diff
@@ -86,37 +103,122 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+            resp.setContentLength((int) agentInstallerJar.length());
+            resp.setDateHeader("Last-Modified", agentInstallerJar.lastModified());
+
+            ZipFile agentInstallerZip = new ZipFile(agentInstallerJar);
+            ZipOutputStream zos = new ZipOutputStream(resp.getOutputStream(), StandardCharsets.UTF_8);
+            final String wildflyHome = getValueFromQueryParam(req, AGENT_INSTALLER_PROPERTY_WILDFLY_HOME, null);
+            final String hawkularServer = getValueFromQueryParam(req, AGENT_INSTALLER_PROPERTY_HAWKULAR_SERVER,
+                    "http://localhost:8080");
+            String defaultModuleDist = hawkularServer.startsWith("http:") ? hawkularServer +
+                    "/hawkular/wildfly-agent/download" : "http://localhost:8080/wildfly-agent/download";
```
jmazzitelli: here, and the line above for wildflyHome - we should be using the hostname of the server (not localhost) because we want the installer to connect to this server to download the agent extension and we want the agent to connect to this server when installed. "localhost" will only work if you are running the installer on the same machine as the kettle instance.

jmazzitelli: I also think you meant for the else clause in the ? : statement to be an https URL (schema=https with port 8443, with /hawkular/ in the path)

modules/hawkular-wildfly-agent-download/src/main/java/org/hawkular/component/wildflymonitor/WildFlyAgentServlet.java
```diff
@@ -86,37 +103,122 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+            final String moduleDist = getValueFromQueryParam(req, AGENT_INSTALLER_PROPERTY_MODULE_DIST,
+                    defaultModuleDist);
+            final String username = getValueFromQueryParam(req, AGENT_INSTALLER_PROPERTY_USERNAME, null);
+            final String password = getValueFromQueryParam(req, AGENT_INSTALLER_PROPERTY_PASSWORD, null);
```
jmazzitelli: very shortly we are going to want to push the offline token stuff - so I would add "securityKey" and "securitySecret" as parameters and put them in the properties file too.

modules/hawkular-wildfly-agent-download/src/main/java/org/hawkular/component/wildflymonitor/WildFlyAgentServlet.java
```diff
@@ -86,37 +103,122 @@ protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws Se
...
+                    zos.putNextEntry(new ZipEntry(AGENT_INSTALLER_PROPERTIES_FILE_NAME));
+                    try (BufferedReader br =
+                                 new BufferedReader(
+                                         new InputStreamReader(agentInstallerZip.getInputStream(entryIn),
+                                                 StandardCharsets.UTF_8)
+                                 )) {
+                        String line;
+                        while ((line = br.readLine()) != null) {
+                            //todo: it'd be better to reuse the strings from the agent installer module to keep it DRY
+                            // and resilient against the changes
```
jmazzitelli: I have code somewhere that can do this kind of thing. I'll dig it out and we can use it here. It doesn't replace commented variables, but it does retain comments in the file. That would at least make this less dependent on things like commented example values of "/opt/wildfly" in the .properties file.

modules/hawkular-wildfly-agent-download/src/main/java/org/hawkular/component/wildflymonitor/WildFlyAgentServlet.java
```diff
@@ -171,23 +273,42 @@ private void copy(InputStream input, OutputStream output) throws IOException {
...
+            if (installerDownloadFile.exists()) {
+                return installerDownloadFile;
+            } else {
+                installerDownloadFile = null; // the file was removed recently - let's look for a new one
+            }
+        }
+
+        File configDir = new File(System.getProperty("jboss.server.config.dir"));
+        for (File file : configDir.listFiles()) {
+            if (file.getName().startsWith("agent-installer") && file.getName().endsWith(".jar")) {
```
jmazzitelli: we want to continue using the official branding 'hawkular-wildfly-agent' not just 'agent'  (because in the world of hawkular, there are many kinds of feeds and agents, hawkular-wildfly-agent being just one flavor.)

So we need to make sure the names of the files are reflective of the hawkular name - "hawkular-wildfly-agent-installer-ABC.jar"

jmazzitelli:
note: do not call this "agent-installer.jar" - use the proper branding "hawkular-wildfly-agent-installer.jar" - "agent-installer" isn't clear what agent you are talking about "is it the vert-x agent? the ruby client? the wildfly agent?" :)

dist/assembly.xml
```diff
@@ -136,6 +136,19 @@
...
     </dependencySet>
 
+    <!-- put the agent installer to the same directory as previous artefact from the same reason -->
+    <dependencySet>
+      <outputDirectory>${hawkular.dist.zip.root.dir}/standalone/configuration</outputDirectory>
+      <useProjectArtifact>false</useProjectArtifact>
+      <includes>
+        <include>org.hawkular.agent:hawkular-wildfly-agent-installer:jar</include>
+      </includes>
+      <outputFileNameMapping>agent-installer.${artifact.extension}</outputFileNameMapping>
```
jmazzitelli: this needs to be called "hawkular-wildfly-agent-installer.${artifact.extension}" - and I would go so far as to put the version string in there so we know exactly what is in there just by looking at the name, but that is less of an issue. But for sure the name needs to be "hawkular-wildfly-agent-installer"

PR #660, opened 2015-11-16
HAWKULAR-806 Removing browse buttons, because they are useles + addin…

Jiri-Kremser:
…g some hints/descriptions + adding an example how to run the agent installer

### DIFF
--- a/console/src/main/scripts/bower.json
+++ b/console/src/main/scripts/bower.json
@@ -37,7 +37,9 @@
     "keycloak": "1.3.1",
     "angular-truncate": "sparkalow/angular-truncate",
     "angular-smart-truncate": "1.0.2",
-    "angular-wizard": "0.5.5"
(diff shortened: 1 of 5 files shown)


console/src/main/scripts/bower.json
```diff
@@ -37,7 +37,9 @@
     "keycloak": "1.3.1",
     "angular-truncate": "sparkalow/angular-truncate",
     "angular-smart-truncate": "1.0.2",
-    "angular-wizard": "0.5.5"
+    "angular-wizard": "0.5.5",
+    "ng-clip": "~0.2.6",
```
mtho11: Just FYI, because of being bitten by changing software packages we have been inclined to use fixed versions and not allow patch versions '~'.

mtho11:
Merged #660

PR #662, opened 2015-11-17
HAWKULAR-752 Persist Date/Time pickers across pages

lucasponce:
@mtho11 or @ammendonca could you please take a look ?
This JIRA is to maintain the datepicker across all pages.
I've tried to not refactor a lot, just using the HawkularNav as a service to maintain this value.

### DIFF
--- a/console/src/main/scripts/plugins/metrics/ts/alertsCenterList.ts
+++ b/console/src/main/scripts/plugins/metrics/ts/alertsCenterList.ts
@@ -27,7 +27,6 @@ module HawkularMetrics {
     public isWorking = false;
     public alertsTimeStart:TimestampInMillis;
     public alertsTimeEnd:TimestampInMillis;
-    public alertsTimeOffset:TimestampInMillis;
     public lastUpdateDate:Date = new Date();
 
     public alertsList:IAlert[];
@@ -47,6 +46,7 @@ module HawkularMetrics {
 
     constructor(private $scope:any,
                 private HawkularAlertsManager:IHawkularAlertsManager,
+                private HawkularNav:any,
                 private ErrorsManager:IErrorsManager,
                 private $log:ng.ILogService,
                 private $q:ng.IQService,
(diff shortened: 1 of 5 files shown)


console/src/main/scripts/plugins/metrics/ts/alertsCenterList.ts
```diff
@@ -57,16 +57,9 @@ module HawkularMetrics {
...
-        $routeParams.timeOffset = 3600000 * 12;
-        $scope.hkParams.timeOffset = $routeParams.timeOffset;
-      }
-
-      this.alertsTimeOffset = $routeParams.timeOffset;
-
       // If the end time is not specified in URL use current time as end time
       this.alertsTimeEnd = $routeParams.endTime ? $routeParams.endTime : Date.now();
-      this.alertsTimeStart = this.alertsTimeEnd - this.alertsTimeOffset;
+      this.alertsTimeStart = this.alertsTimeEnd - HawkularNav.getTimeOffset();
```
mtho11: I have some doubts with this PR. It looks to bypass the url offset parameters which give each screen its stateless RESTful url architecture. Each screens data keys should be available from the url and stored somewhere.

lucasponce: Point is that if we don't have state, url should be rewritten with the timeoffset parameters.
Or perhaps I miss some alternative ?

ammendonca: @lucasponce, In case there's a timeOffset already set in the URL, that what we use.. if there isn't any, we use the default (now 12h) as in https://github.com/hawkular/hawkular/blob/6eb079e181ec7cfd98974c57bde65a85b3fb53a6/console/src/main/scripts/plugins/directives/topbar/ts/topbarServices.ts#L47L58.

We should rely on hkParams.timeOffset and always provide our links making use of it. We are losing the range selection when moving between top level pages, eg, the Alerts Center link at https://github.com/hawkular/hawkular/blob/6eb079e181ec7cfd98974c57bde65a85b3fb53a6/console/src/main/scripts/plugins/directives/topbar/html/topbar.html#L3 should be `<a href="/hawkular-ui/alerts-center/{{hkParams.timeOffset}}">Alert Center</a>`

lucasponce:
@mtho11, @ammendonca thx for the review ! I wasn't sure of it as well, so it's great you confirmed the suspects. I'm going to use Alex's suggestion to introduce the hkParams on the url to see if there is not additional changes on it. I'm going to close this PR.

PR #666, opened 2015-11-18
HAWKULAR-414 Datasource wait time color is too light

mtho11:
Also added some additional type safety for ResourceId in the files that were touched.


mtho11:
Merged #666

PR #667, opened 2015-11-18
HAWKULAR-812 - Use AES/CBC/PKCS5PADDING for agent encryption with a PBKDF2 key.

jpkrohling:


jpkrohling:
@jmazzitelli , another one for you :)

modules/hawkular-wildfly-agent-download/src/main/java/org/hawkular/component/wildflymonitor/WildFlyAgentServlet.java
```diff
@@ -267,22 +287,18 @@ private void downloadAgentInstaller(HttpServletRequest req, HttpServletResponse
      *
      * @param properties the map where the property is found; this will be updated with the newly encoded value
      * @param propertyName the name of the property that needs to be encoded
-     * @param encryptionKey the key to use when encoding the value
      */
-    private void encode(HashMap<String, String> properties, String propertyName, String encryptionKey)
+    private void encode(HashMap<String, String> properties, String propertyName, Cipher cipher, SecretKeySpec keySpec)
             throws Exception {
         String clearText = properties.get(propertyName);
-        if (clearText != null) {
```
jmazzitelli: we need to keep the if-stmt here. if the clearText is null, that means the user never specified the value in the first place, so we do not want to put a value in the hashmap "properties".

PR #668, opened 2015-11-18
HAWKULAR-752 Propagate timeOffset params across views

lucasponce:
@mtho11, @ammendonca I've propagated the hkParams on urls where datepickers are used.
There is a pending question, if we should propagate hkParams across all views.
This PR only works on the case between alerts center, url, and app-server details, which I think is the most visible.
Please, let me know how it looks like, or feel free to take it as start point if you want to add some changes on top.

### DIFF
--- a/console/src/main/scripts/plugins/directives/subtab/html/subtab.html
+++ b/console/src/main/scripts/plugins/directives/subtab/html/subtab.html
@@ -2,12 +2,12 @@
   <div>
     <div class="row hk-top-row">
       <div class="col-xs-7">
-        <a ng-show="isAppServerPage()" href="/hawkular-ui/app/app-list" class="back">« All Application Servers</a>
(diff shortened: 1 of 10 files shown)


lucasponce:
Adding a small fix for HAWKULAR-811.
Karel, please, could you review as well ?
(Sorry, I can't find your username to link the notification)

mtho11:
LGTM. wdyt @ammendonca, Karel

ammendonca:
It looks good to me too. I think we can re-work/think it later, some pages (like url list, app server list) are not bound to a time range, thus passing it may seem strange. In theory once we have it on the rootScope, unless we override it in some way, it should persist across navigation.. but agree this is the best option for now, guarantees it gets persisted.

mtho11:
Merged #668

PR #672, opened 2015-11-19
HAWKULAR-127

karelhala:
Change version of angular-bootstrap and angular-patternfly so we can use newest version of simple filter from angular patternfly. Create new enums ServerStatus - for easy creation of server statuses. Server type - egain easy creation of server type for filtering. Create new directive, TableFilters for filtering over specific table. insert this directive to server list. Fix of NaN in Web alerts. Add search function to controller, which takes care of data. This way we can use searching in every table

### DIFF
--- a/console/src/main/scripts/plugins/metrics/ts/directives/filterDirective.ts
+++ b/console/src/main/scripts/plugins/metrics/ts/directives/filterDirective.ts
@@ -1,0 +1,101 @@
+///
+/// Copyright 2015 Red Hat, Inc. and/or its affiliates
+/// and other contributors as indicated by the @author tags.
+///
+/// Licensed under the Apache License, Version 2.0 (the "License");
+/// you may not use this file except in compliance with the License.
+/// You may obtain a copy of the License at
+///
+///    http://www.apache.org/licenses/LICENSE-2.0
+///
+/// Unless required by applicable law or agreed to in writing, software
+/// distributed under the License is distributed on an "AS IS" BASIS,
+/// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+/// See the License for the specific language governing permissions and
+/// limitations under the License.
+///
+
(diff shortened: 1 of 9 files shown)


console/src/main/scripts/plugins/metrics/ts/directives/filterDirective.ts
```diff
@@ -0,0 +1,101 @@
...
+      };
+    }
+  }
+
+  let filterLink = ($scope:any, element:ng.IAugmentedJQuery, attrs:ng.IAttributes) => {
+    $scope.vm.initFilters($scope.activeFilters, $scope);
+    $scope.vm.watchResourceList($scope);
+  };
+
+  export class HkTableFilters {
```
mtho11: Directives should end in *Directive so we can easily find them in the ide by just search *.Directive. And filters should end in *Filter.

console/src/main/scripts/plugins/metrics/ts/metricsAppServerList.ts
```diff
@@ -21,6 +21,35 @@
 
 module HawkularMetrics {
 
+  export class ServerStatus {
```
mtho11: Maybe this should be in its own file as its generally useful in many places 'serverStatus.ts'

console/src/main/scripts/plugins/metrics/ts/metricsAppServerList.ts
```diff
@@ -60,10 +92,71 @@ module HawkularMetrics {
         $rootScope.$watch('currentPersona', (currentPersona) => currentPersona &&
         this.getResourceList(currentPersona.id));
       }
+      this.serverStatusArray = Object.keys(ServerStatus).map(function(type) {
```
mtho11: Let's use the ES6 fat arrow ()=> notation instead of old 'function()'

mtho11:
LGTM. Merged #672

PR #675, opened 2015-11-19
Hawkular-464 Deployment Failure Event Alerting

jshaughn:
Add alert generation for failed deployments.
- New src-deps versions for component enhancements to support this
  - with mazz, enhanced bus-commons with better classloading/deserialization support for MDBs
  - with mazz, enhanced command-gateway to support multiple commands on ws responses, now can notify UI and put a response on the bus
  - enhanced alerts to consume ws deploy app response messages on bus and generate events
- Added new trigger definition on app-servers to fire on a failed deployment event
  - note, also storing successful deployment events for future use/display
- Added new UI support for 'Event' trigger-type and DEPLOYMENT_FAIL alert-type

Also:
- fixed time-window bug in getAlerts query for some trigger types
- make bus-commons dep provided (by nest) for pinger
- make bus-commons dep provided (by nest) for avail-creator

### DIFF
--- a/console/src/main/scripts/plugins/metrics/ts/alertsCenterTriggerList.ts
+++ b/console/src/main/scripts/plugins/metrics/ts/alertsCenterTriggerList.ts
@@ -188,17 +188,24 @@ module HawkularMetrics {
       let route = 'unknown-trigger-type';
       let encodedId = this.encodeResourceId(trigger.id);
 
-      if ('Availability' === trigger.context.triggerType) {
-        route = '/hawkular-ui/alerts-center-triggers/availability/' + encodedId;
-      } else if ('Event' === trigger.context.triggerType) {
-        route = '/hawkular-ui/alerts-center-triggers/event/' + encodedId;
-      } else if ('Range' === trigger.context.triggerType) {
-        route = '/hawkular-ui/alerts-center-triggers/range/' + encodedId;
-      } else if ('RangeByPercent' === trigger.context.triggerType) {
-        route = '/hawkular-ui/alerts-center-triggers/range-percent/' + encodedId;
-      } else if ('Threshold' === trigger.context.triggerType) {
-        route = '/hawkular-ui/alerts-center-triggers/threshold/' + encodedId;
+      switch (trigger.context.triggerType) {
+        case 'Availability' :
+          route = '/hawkular-ui/alerts-center-triggers/availability/' + encodedId;
+          break;
+        case 'Event' :
+          route = '/hawkular-ui/alerts-center-triggers/event/' + encodedId;
+          break;
+        case 'Range' :
+          route = '/hawkular-ui/alerts-center-triggers/range/' + encodedId;
+          break;
+        case 'RangeByPercent' :
+          route = '/hawkular-ui/alerts-center-triggers/range-percent/' + encodedId;
+          break;
+        case 'Threshold' :
+          route = '/hawkular-ui/alerts-center-triggers/threshold/' + encodedId;
+          break;
       }
+
       return route;
     }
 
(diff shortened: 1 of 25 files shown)


console/src/main/scripts/plugins/metrics/ts/app-details/appServerDatasourcesDetails.ts
```diff
@@ -240,6 +224,27 @@ module HawkularMetrics {
       this.getDrivers();
     }
 
+    private getAlerts(res:any):void {
```
mtho11: Would be nice to have the type here.

console/src/main/scripts/plugins/metrics/ts/app-details/appServerDatasourcesDetails.ts
```diff
@@ -240,6 +224,27 @@ module HawkularMetrics {
...
+      let dsArray:IAlert[];
+      let promise = this.HawkularAlertsManager.queryAlerts({
+        statuses: 'OPEN',
+        tags: 'resourceId|' + res.id,
+        startTime: this.startTimeStamp,
+        endTime: this.endTimeStamp
+      }).then((data)=> {
```
mtho11: The type would be good here too as anyone outside of alerts wouldn't know what kind of data is in 'data'. The types let our IDE's autocomplete this stuff.

console/src/main/scripts/plugins/metrics/ts/app-details/appServerDatasourcesDetails.ts
```diff
@@ -240,6 +224,27 @@ module HawkularMetrics {
...
+        _.forEach(data.alertList, (item) => {
+          item['alertType'] = item.context.alertType;
```
mtho11: It is preferable to type the item(IAlert) then use item.alertType instead of item['alertType'] which is not typed. This just helps to cuts down on typos and stupid little errors in the long run.

console/src/main/scripts/plugins/metrics/ts/app-details/appServerDatasourcesDetails.ts
jshaughn: Yes, this is actually not new, it's an artifact of the original code that I've been working on.  Same for the 'data' below.  I can look to see if I can improve it...

console/src/main/scripts/plugins/metrics/ts/app-details/appServerDeploymentsDetails.ts
```diff
@@ -57,7 +58,8 @@ module HawkularMetrics {
       $scope.vm = this;
       HawkularOps.init(this.NotificationsService);
 
-      this.startTimeStamp = +moment().subtract(1, 'hours');
+      this.resourceId = this.$routeParams.resourceId;
+      this.startTimeStamp = +moment().subtract(($routeParams.timeOffset || 3600000), 'milliseconds');
```
mtho11: We need to handle this default time period a bit better, so that if we change it in one place it gets changed everywhere. Nothing you need to do here just making a note of this. Ultimately, it should probably be user-definable anyway.

console/src/main/scripts/plugins/metrics/ts/app-details/appServerJvmDetails.ts
```diff
@@ -126,36 +146,16 @@ module HawkularMetrics {
...
-      this.getAlerts(this.$routeParams.resourceId, this.startTimeStamp, this.endTimeStamp);
-    }
-
     public getJvmData():void {
-      this.endTimeStamp = this.$routeParams.endTime || +moment();
-      this.startTimeStamp = this.endTimeStamp - (this.$routeParams.timeOffset || 3600000);
-
       this.MetricsService.retrieveGaugeMetrics(this.$rootScope.currentPersona.id,
-        'MI~R~[' + this.$routeParams.resourceId + '~~]~MT~WildFly Memory Metrics~Heap Used',
+        'MI~R~[' + this.resourceId + '~~]~MT~WildFly Memory Metrics~Heap Used',
```
mtho11: Why the change away from $routeParams all screens should be driven from the $routeParams to make our screens stateless and RESTful.
The url to a screen is like an API, one should be able to change the resourceId (or any other params in the uri) and the screen should change to reflect the new uri.

console/src/main/scripts/plugins/metrics/ts/metricsAppServerDetails.ts
```diff
@@ -541,11 +541,64 @@ module HawkularMetrics {
...
+        .existTrigger(this.resourceId + '_failed_deployment').then(() => {
+          this.$log.debug('Failed Deployment trigger exists, nothing to do');
+        }, () => {
+          // Failed Deployment trigger doesn't exist, need to create one
+
+          let triggerId:string = this.resourceId + '_failed_deployment';
+          let resourceId:string = triggerId.slice(0, -18);
+          let dataId:string = this.resourceId + '_DeployApplicationResponse';
+
+          let fullTrigger = {
```
mtho11: Add typing around fullTrigger. 
Replace with 'let fullTrigger:IAlertTrigger = {'

console/src/main/scripts/plugins/metrics/ts/alertsCenterTriggerList.ts
```diff
@@ -190,6 +190,8 @@ module HawkularMetrics {
 
       if ('Availability' === trigger.context.triggerType) {
         route = '/hawkular-ui/alerts-center-triggers/availability/' + encodedId;
+      } else if ('Event' === trigger.context.triggerType) {
```
mtho11: Switch might be easier to maintain?

console/src/main/scripts/plugins/metrics/ts/app-details/appServerJvmDetails.ts
jshaughn: In the constructor we have:

  this.resourceId = this.$routeParams.resourceId;

I just used the local var instead of going through routParams everywhere. Is that not OK?

mtho11: No, otherwise this is bypassing the RESTful params.

jshaughn: I'll change it back but I don't really get it, I don't know then what the point is of a variable scoped at the class level.  Are you saying we want to support hand-editing of the url without actually having the class get re-constructed?

mtho11: @jshaughn sorry you are correct about:'this.resourceId = this.$routeParams.resourceId;' being equivalent. When you are just looking through the diffs and you see a change from *$routeParams.resourceId* to *resourceId* it looks like a bad change ;)

mtho11: For more self describing code: the *$routeParams.resourceId* is clear where it came from whereas *resourceId* is not clear that it came from the url (you have to look into where it was assigned somewhere else in the code).

mtho11:
Merged #675

PR #680, opened 2015-11-24
Update Bower to version 1.6.3.

ammendonca:
It seems something started causing issues with non-numeric bower versions.


mtho11:
Merged #680

PR #681, opened 2015-11-24
HAWKULAR-841 : Show XA Datasources in the DS tab

ammendonca:
XA Datasources are separated from non-XA Datasources,
requiring independent queries to inventory to retrieve them.


mtho11:
Merge #681

PR #683, opened 2015-11-25
HAWKULAR-843 : Fix dropdowns across console

ammendonca:
There were changes to angular-ui bootstrap which caused the dropdowns to stop working.
The dropdown top-most element requires a "dropdown" directive, as class="dropdown"
is not sufficient.


mtho11:
Nice work Alex!

PR #685, opened 2015-11-26
Alert Routering

karelhala:
Add alert router, which is responsible for registering functions when alerts are recieved from server. Register to this router Datasources, Deployments, JVM and Web. Remove unnnecessary inject field from these files and remove unused variables.


mtho11:
LGTM!

PR #688, opened 2015-11-27
Enable file deletion only if build not in progress

ammendonca:
File deletion may cause trouble, as multiple changes will trigger simultaneous builds,
which often end up causing gulp to fail (exit) due to trying to operate on deleted files.

With this change the delete only happens if a build is not in progress already.


mtho11:
Merged #688

PR #691, opened 2015-12-01
Replace truncate libraries with single custom filter

ammendonca:
Filter usage:
{{ "some text" | truncate : <length> : <words> : <left|right|middle> | <separator> }}

- length: the truncated size (defaults to 15)
- words: keep words intact (defaults to false)
- left|right|middle: where to truncate (defaults to left)
- separator: the separator char/string (defaults to '...')


mtho11:
Merged #691

PR #693, opened 2015-12-03
For Bower builds, use git SHA# instead of branches (essentially snapsh…

mtho11:
…ots)  during dev periods. This should eliminate any dependent projects from breaking the Hawkular project (like a bad version of hawkular-charts or hawkular-ui-services).

A day or two before Milestone release we will still use regular bower versions to tighten up even more on approved changes going into the release (like we have been doing right around release).


mtho11:
Merged #693

PR #695, opened 2015-12-04
Adapt UI to Agent and Metrics IDs change

ammendonca:
Update inventory to 9.0.Final too.

### DIFF
--- a/console/src/main/scripts/plugins/metrics/ts/services/metricsService.ts
+++ b/console/src/main/scripts/plugins/metrics/ts/services/metricsService.ts
@@ -91,6 +91,18 @@ module HawkularMetrics {
                 private NotificationsService:INotificationsService) {
     }
 
+    /**
+     * Generates metric Id given the components
+     * @param type the metric type, as a char: 'A'vailability or 'M'etric
+     * @param feedId the feed Id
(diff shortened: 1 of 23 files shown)


console/src/main/scripts/plugins/metrics/ts/services/metricsService.ts
```diff
@@ -91,6 +91,18 @@ module HawkularMetrics {
...
 
+    /**
+     * Generates metric Id given the components
+     * @param type the metric type, as a char: 'A'vailability or 'M'etric
+     * @param feedId the feed Id
+     * @param resId the resource Id
+     * @param metricId the metric Id
+     * @returns {string} a string uniquely identifying the requested metric
+     */
+    public static getMetricId(type: string, feedId: FeedId, resId: ResourceId, metricId: MetricId): string {
```
mtho11: There is a ResourceType available instead of string.

ammendonca: Yeah.. but that's actual a char 'A' or 'M'.. it doesn't fit that well there. That's more of a metric type than resource type. We can certainly improve on that later

mtho11:
Merged #695

PR #697, opened 2015-12-04
HAWKULAR-178 - Last used persona is the default upon login

jpkrohling:


jpkrohling:
For this to properly work, the following Accounts version needs to be used in Hawkular. As that version is already on Wildfly 10, it cannot be used.

This code, however, is prepared to fallback to the current behavior in such case, so, no updates to accounts are required for this to be merged.

jpkrohling:
Accounts version that this feature will start to work as intended is: 2.0.7.Final-SRC-60d1a19656198451d44a429bc8646cdb48ad89da

jpkrohling:
@mtho11 , could you please review ?

console/src/main/scripts/plugins/accounts/ts/persona.ts
```diff
@@ -49,12 +49,42 @@ module HawkularAccounts {
 
     public loadData():void {
       this.loading = true;
-      this.currentPersona = this.HawkularAccount.Persona.get({id: 'current'},
+      let personaToGet = 'current';
+
+      let lastPersonaId = localStorage.getItem('lastPersona');
```
mtho11: Aha localStorage. Good to know its being used. QE might also want to be aware of this.

mtho11:
LGTM. @jpkrohling will you add the Accounts version to the pom for this PR as well?

console/src/main/scripts/plugins/accounts/ts/persona.ts
jpkrohling: Why? :)

jpkrohling:
Not really, this can be merged without the update to Accounts, as it will just fallback to the default behavior. Once the WF10 task is finished, a newer version of Accounts will be used, so, this will work 'automatically'.

console/src/main/scripts/plugins/accounts/ts/persona.ts
mtho11: For setting up different default users. Perhaps not so interesting.

mtho11:
Merged #697

PR #702, opened 2015-12-09
Use resources "name" instead of playing with "id"

ammendonca:
With the renaming made in the agent, it is now possible/easier to just
use the name instead of playing around with trimming the id or the name.

Also update to latest commit of ui-services required by Update DS.


mtho11:
Merged #702 
The itests aren't passing but that is another issue entirely.

PR #704, opened 2015-12-11
New module for common Hawkular REST api - first shot

Jiri-Kremser:
It's a new war living under the `/modules/`

Currently it's only capable of CRUDing the URLs with all the magic behind it (except pre-creation of alerts, this will be the next step)

It uses the [Hystrix](https://github.com/Netflix/Hystrix) library and java rx, so each operation is basically encapsulated as a Hystrix command with possible fallbacks if the backend is not there. Hystrix inherently supports the [Circuit breaker pattern](https://github.com/Netflix/Hystrix/wiki/How-it-Works#CircuitBreaker), so it will be possible to fallback to some dummy response if the system is overloaded and other interesting features.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,11 @@ install:
 - mvn -s .travis.maven.settings.xml -version -B
 - df -h
 - du -sh $HOME/.m2/repository
+
+before_script:
+- ulimit -u 4096
+- ulimit -s 65535
+- ulimit -a
 script:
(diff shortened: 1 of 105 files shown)


mtho11:
Perhaps this should be squash merged.

Jiri-Kremser:
> Perhaps this should be squash merged.

I've squhashed everything to 1 big commit

PR #705, opened 2015-12-11
HAWKULAR-586

karelhala:
add new page server overview. Add all alerts to this page, add graph about all alert in timeline, add graph about avarege JVM, add graph about connections to DB, add graph about web connections, add information about deployments. Show basic server information. Create labels inside overview panel. Labels can be added and removed. Fix graph with alerts, not it is much understandale and it show actual alerts in time. Fix NaN in web session alerts.

### DIFF
--- a/console/src/main/scripts/plugins/metrics/less/metrics.less
+++ b/console/src/main/scripts/plugins/metrics/less/metrics.less
@@ -43,6 +43,17 @@ body {
   float: none;
 }
 
+.label-editor-dropdown {
+  width: 400px;
+  padding: 15px;
+}
+
+.server-label {
+  display: inline-block;
+  padding-right: 5px;
+  cursor:pointer;
+}
+
 .container {
   @media (min-width: 1201px) and (max-width: 1440px) {
     width: calc(~"100% - 40px");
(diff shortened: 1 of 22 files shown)


console/src/main/scripts/plugins/metrics/less/metrics.less
```diff
@@ -2458,6 +2469,7 @@ section {
 
   padding-bottom: @grid-gutter-width/2;
   padding-top: @grid-gutter-width/2;
+  height: 120px;
```
mtho11: Hmm, I'm little worried about this height being applied to .container

cardosogabriel: Actually it applies to .hk-dashboard-block {}
@karelhala, did you test in different resolutions (from 768px width above)? We usually don't define heights like this since the content adjust. For this situation we should use a plugin like https://github.com/liabru/jquery-match-height (which I used in my prototype), which makes the elements of the same line with the same height.

mtho11:
LGTM. Merged #705

console/src/main/scripts/plugins/metrics/less/metrics.less
karelhala: Well to be honest I did not tested it on different resolutions. I just didn't liked the way it snapped to new line, it was some kind of hot fix. So please @cardosogabriel if it is bad thing just give me a tip what is better, thanks.

PR #709, opened 2015-12-11
Pull in latest charts and make bower versions static for release.

mtho11:
Pulls in charts version with hovers Gabriel cleaned up and a sparkline fix for when values are 0 they render below the x-axis


mtho11:
Merged #709

PR #712, opened 2015-12-14
Better fix to Experimental Mode tabs enable/disable

ammendonca:
Not using fixed index as it can easily get broken, using instead a list of experimental tabs.


mtho11:
Merged #712

PR #718, opened 2015-12-15
HAWKULAR-815

karelhala:
change the URLs to be shown as tiles. Show sparkline graph at bottom of each tile. Style the URL detail with sidebar alerts. Change the view of filter, to be at same line.


mtho11:
Merged #718

PR #729, opened 2015-12-17
HWKAGENT-45 allow user to define local managed server name in installer config

jmazzitelli:


jmazzitelli:
The JIRA referenced by the branch name is incorrect. This really belongs to HWKAGENT-45.

PR #730, opened 2015-12-17
Hawkular-464 Fix regression in original feature work

jshaughn:
- improve inline documentation
- update alerts srcdep to version needed for fix

PR #732, opened 2015-12-17
Hawkular-887 NaN alert values

jshaughn:
The fix for this may also improve accuracy for non-NaN values because we
weren't handling 0 values correctly.

### DIFF
--- a/console/src/main/scripts/plugins/metrics/ts/services/alertsManager.ts
+++ b/console/src/main/scripts/plugins/metrics/ts/services/alertsManager.ts
@@ -363,12 +363,12 @@ module HawkularMetrics {
                   }
                 }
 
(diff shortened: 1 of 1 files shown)


console/src/main/scripts/plugins/metrics/ts/services/alertsManager.ts
```diff
@@ -363,12 +363,12 @@ module HawkularMetrics {
                   }
                 }
 
-                if ( evalItem.rate ) {
+                if ( undefined != evalItem.rate ) {
```
mtho11: I'm pretty sure that *if( !evalItem )* will get the falsy condition of undefined

mtho11:
Merged #730

PR #734, opened 2015-12-17
HAWKULAR-519 : Show operation in progress spinner for deployment actions

ammendonca:
Also fix an issue with ng-show/hide and fa-spin and re-enable debug info


mtho11:
Merged #734

PR #739, opened 2015-12-18
Upgrade to WildFly 10

ppalaga:


ppalaga:
In 5e133fb the itests are passing but the UI is not there yet.

ppalaga:
5d7c427 should be close to what we need, but the tests are failing.

jpkrohling:
The following branch has additional fixes for this PR:
https://github.com/jpkrohling/hawkular/tree/jpk-151217-wf10

Known issues: 

* `keycloak.import` is not being overwritten during `-Pdev`
* `pom.xml` is currently referencing metrics-snapshot, will need to be changed to the src-deps once the following gets merged: hawkular/hawkular-metrics#419

jpkrohling:
The last item is fixed (thanks @jmazzitelli !). I also rebased/squashed.

jpkrohling:
First item fixed as well: keycloak.import is now being correctly set. Also, the agent is being marked with "enabled=false" for production builds, so that it won't attempt to send metrics as jdoe/password.

jmazzitelli:
@jpkrohling I would add some additional things here for dev builds. We really should only set the jdoe/password if its a dev build. Anything else, we should put their values to  "SET_ME" to indicate to the user that before they enable the agent they should be set (in the future, these will be security-secret and security-key). So in the dev profile, you can have
```xml
<agent.enabled>true</agent.enabled>
<agent.username>jdoe</agent.username>
<agent.password>password</agent.password>
```

with those set at the top also but values of "SET_ME" for both:

```xml
<properties>
  <agent.enabled>false</agent.enabled>
  <agent.username>SET_ME</agent.username>
  <agent.password>SET_ME</agent.password>
  ...
```

Then in exec plugin config you just pass them in:

```xml
<argument>--username=${agent.username}</argument>
<argument>--password=${agent.password}</argument>
```

jpkrohling:
+1, the latest state of my branch has now the behavior that @jmazzitelli described.

ppalaga:
About 61ecb68 :

* it uses the work done by @jpkrohling  in https://github.com/jpkrohling/hawkular/tree/jpk-151217-wf10
* it upgrades some hawkular components to versions that were released in between
* it fixes `jboss-deployment-structure.xml` of metrics war to depend on bus rather than nest WF module

This PR is good to merge from my PoV now.

@jpkrohling could you please look through it and eventually merge?

ppalaga:
c893306 fixes the date in the top level pom.xml of 61ecb68

jpkrohling:
Looks good to me, and it can be merged once Travis approves it.

ppalaga:
c893306 is yet another attempt to fix the license header properly

PR #741, opened 2015-12-21
HAWKULAR-896 : Respect the filter when selecting all alerts/triggers

ammendonca:
Also fixes HAWKULAR-828 where the checkbox wasn't checked when clicking on the cell


mtho11:
Merged #741

PR #742, opened 2015-12-22
HAWKULAR-803 Prelim - first commit of context chart until fully respo…

mtho11:
…nsive and metrics capabiltiy to return start of metrics collection period.

Chart part  commented out until other things are resolved with metrics colleciton period and fully responsive charts that aren't reliant upon  different aspect ratios to allow full alignment at various responsive sizes.


mtho11:
Merged #742

PR #753, opened 2015-12-30
HAWKULAR-902 : Increase toastr timeout to 8 seconds

ammendonca:
Also enabled the option to prevent opening duplicates of last toastr


mtho11:
Merged #753

PR #757, opened 2016-01-05
Adapt gulp-watch dist target to new location

ammendonca:
The new dist hawkular path has one less hawkular-<version> directory


mtho11:
Merged #757

PR #758, opened 2016-01-06
HAWKULAR-864

karelhala:
Add percent number in donut graphs. Fix labels, so it is not needed to wait for refresh after inserting first label. Remove number of results in server list. Change message when no results are filtered. Fix wrong number of visible results, when filtering (filtering now on javascript not in HTML).


mtho11:
Merged #758

PR #760, opened 2016-01-07
topology-view

Jiri-Kremser:
* fixing the ugly tooltips (css)
* fetching the a configuration for servers (feed id was HCoded for server nodes)
* fixing the links when double clicking on the items (the urls in the router were changed recently)


mtho11:
Merged #760

PR #772, opened 2016-01-14
Bumping inventory to 0.12.1.Final.

metlos:


ppalaga:
@metlos it is nice that you care, but it is not always safe to upgrade Inventory without upgrading the rest, i.e. Commons and Accounts that your Inventory version depends on and cmdgw, Alerts and Agent that depend on Invetory.

ppalaga:
... and hawkular-parent

metlos:
Ok, leaving it to more capable hands.

PR #775, opened 2016-01-15
Es6 gulpfile

mtho11:
Get all the ES6 goodness in a our gulpfile as now as well.


mtho11:
Hmm, still issue with License check even after adding the exclusion to the POM.

[INFO] Checking licenses...
[WARNING] Unknown file extension: /home/travis/build/hawkular/hawkular/console/src/main/scripts/.babelrc

@ammendonca or @ppalaga any suggestions?

ammendonca:
Yeah, what you've changed is relative to checkstyle and not license, license exclusions are below @ https://github.com/hawkular/hawkular/blob/es6-gulpfile/console/pom.xml#L133-L137

mtho11:
D'oh! Checkstyle is not maven license checker. Thanks @ammendonca!

mtho11:
Merged #775

PR #780, opened 2016-01-15
hot-fixing the install agent page 

Jiri-Kremser:
Currently it's broken and it's downloading the module zip file instead of the installer jar, because the server-side was slightly changed.

This change:
* changes the url that was changed in the servlet (`/hawkular/wildfly-agent/download?installer=true` -> `/hawkular/wildfly-agent/installer`)
* removes the 'Required' form section and makes everything optional
* adds the copy button that will copy the `java -jar foo..` command into user's clipboard
* removes the version information that is no longer valid

NOTE1: this page will be probably different in the future this is a hot fix to have something working for the next alpha

NOTE2: I wasn't able to make the [zeroclipboard](https://github.com/zeroclipboard/zeroclipboard) + [ngClip](https://github.com/asafdav/ng-clip) working when [this](https://git.io/vzcDz) was set to false, so I used the [angular-clipboard](https://github.com/omichelsen/angular-clipboard). ZeroClipboard depends on Flash but it's pretty much the current standard (used by GitHub for instance) and works with old browsers. The angular-clipboard is pretty new and uses the new features in modern browsers, so it doesn't require the Flash, but quite a new browser (Chrome 43+, Firefox 41+, Opera 29+ and IE10+)

### DIFF
--- a/console/src/main/scripts/bower.json
+++ b/console/src/main/scripts/bower.json
@@ -24,7 +24,7 @@
     "angular-toastr": "1.6.0",
     "animate.css": "3.0.0",
     "ngInfiniteScroll": "1.2.1",
-    "hawkular-charts": "hawkular/hawkular-charts#6610b0d",
+    "hawkular-charts": "hawkular/hawkular-charts#6610b0d82706518b22c5f89c3c6e87a0fd22ceeb",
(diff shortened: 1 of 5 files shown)


mtho11:
LGTM

mtho11:
Merged #780

PR #783, opened 2016-01-16
Server overview reformat

mtho11:
Align similar  style charts on same row and add x and y axes to the chart for better context.


mtho11:
Merged #783

PR #785, opened 2016-01-18
HAWKULAR-944 : Remove the decimal part from the GC time alert text

ammendonca:


console/src/main/scripts/plugins/metrics/html/directives/alert-summary.html
```diff
@@ -23,7 +23,7 @@
     <strong>Non Heap Memory</strong>: The Non Heap memory usage was {{alert.avg / 1024 / 1024 | number:2}} MB.
   </div>
   <div ng-switch-when="GARBA">
-    <strong>Garbage collection</strong>: The GC averaged more than {{alert.avg | number:2}} ms per minute.
+    <strong>Garbage collection</strong>: The GC averaged more than {{alert.avg | number:0}} ms per minute.
```
jpkrohling: Is it possible to change the message to adapt to the size of the alert.avg? For instance, if it's higher than 1000ms, then report in seconds with up to 2 decimal places. If it's higher than 60s, report it in minutes, and so on.

I'd prefer to read "2.3s" as opposed to "2,300.00ms".

mtho11: That is probably a separate jira to find lowest common time measure. It  should also be done for bytes and just numbers (could be other UOM as well).

mtho11:
Merged #785

console/src/main/scripts/plugins/metrics/html/directives/alert-summary.html
jpkrohling: That was actually the target of the JIRA for this commit, but in any case, I created a new one: HAWKULAR-946

PR #787, opened 2016-01-19
Update Patternfly to 2.8.0 and have angular-patternfly loaded after it

ammendonca:


jpkrohling:
This fixes the issue for me.

PR #792, opened 2016-01-19
HAWKULAR-951 : Limit decimal places to 1 in DS Pie Chart.

ammendonca:
Also fixes incorrect percentages shown in the pie hover


mtho11:
Merged #792

PR #795, opened 2016-01-20
HAWKULAR-935 : Wait for $rootScope.currentPersona to exist @ DS Detail

ammendonca:
Also some misc fixes:
- Rename typo in method name getDatasource();
- Show the "»" after "All Application Servers" only when there's some entry after;


mtho11:
Merged #795

PR #796, opened 2016-01-21
HAWKULAR-927

karelhala:
add Gabriel's new metrics.less file. Change overview and fix alignment of elements on overview.


mtho11:
Merged #796

PR #798, opened 2016-01-22
HAWKULAR-964 : Fix toggling individual metrics in Platform FS Charts

ammendonca:
Also corrects some text and labels in the FS Charts


mtho11:
Merged #798

PR #805, opened 2016-01-26
HAWKULAR-949 - Refactored the Tokens UI.

jpkrohling:
- Tokens UI is now using cards
- Possibility to add a name/description to a token
- Possibility to set an expiration date
- Display QR Code for a token


mtho11:
Merged #805 
QR codes, cool!

PR #806, opened 2016-01-26
HAWKULAR-980 : Show drivers in Edit DS wizard only when retrieved

ammendonca:
This is related to an issue with pfSelect not updating when ng-options change, thus we only render when we have the data.
angular-patternfly issue @ https://github.com/patternfly/angular-patternfly/issues/177


mtho11:
Merged #806

PR #807, opened 2016-01-26
HAWKULAR-981 Add gcc 4.8 so the node.js node-gyp files can compile.

mtho11:


jmazzitelli:
Mike says he promises that its OK that the travis build is red and that I am to be left unaccountable even if I merge. So, blame him for world chaos if this merge breaks something. :)

PR #808, opened 2016-01-27
HAWKULAR-904

karelhala:
fix select in trigger setup. Also I have noticed that overview and datasources were not auto refreshed and had autorefresh commented out. So I turned auto refresh for these parts.


mtho11:
Merged #808

PR #809, opened 2016-01-27
Clean the directories cached by Travis once their size exceeds the space left on

ppalaga:


ppalaga:
@jpkrohling could you please review and eventually merge?

jpkrohling:
The logic makes sense to me, but I have no idea whether it would work on Travis for real :) So, I'll merge, and let's see if Travis likes it.

PR #810, opened 2016-01-27
Clean up travis.yml

ppalaga:


mtho11:
Merging even though itests are failing, but that is another issue.

PR #811, opened 2016-01-27
HAWKULAR-986 : Remove "Export JDR" button from App Server tabs template

ammendonca:
Since the server overview already has the operation in the proper place, removing the temporary solution.


jpkrohling:
Does this removes the button from all the screens?

ammendonca:
Yes, this is the "master template" for App server tabs.

jpkrohling:
Perfect then!

mtho11:
Merged #811

PR #812, opened 2016-01-28
Run travis itests against an external Cassandra

ppalaga:


tsegismont:
I know the PR is not finished yet but I have a comment on cleaning the C* cache directory. Are you sure it is useful? You will need the archive at the beginning of the build to install Cassandra. And the `.travis.install.cassandra.sh` script takes care of keeping just one version already.

ppalaga:
> I know the PR is not finished yet but I have a comment on cleaning the C* cache directory. Are you sure it is useful? You will need the archive at the beginning of the build to install Cassandra. And the .travis.install.cassandra.sh script takes care of keeping just one version already.

I suppose, you mean this https://github.com/ppalaga/hawkular/commit/1d4ab6ac58cf2d3742fae6e82d6ebaf2e083daad#diff-354f30a63fb0907d4ad57269548329e3R57 ? If you mean that we should remove only `"$HOME/.m2/repository"` that's perhaps a good idea, given that it is only `"$HOME/.m2/repository"` that tends to grow over time. I neglected that because the cleaning of the cache happens only from time to time, IIRC for Hawkular, it is several weeks now.

tsegismont:
Le 28/01/2016 12:57, Peter Palaga a écrit :
> I suppose, you mean this
> ppalaga@1d4ab6a#diff-354f30a63fb0907d4ad57269548329e3R57
> <https://github.com/ppalaga/hawkular/commit/1d4ab6ac58cf2d3742fae6e82d6ebaf2e083daad#diff-354f30a63fb0907d4ad57269548329e3R57>

Yes

> ? If you mean that we should remove only |"$HOME/.m2/repository"| that's

Yes. The install scripts already cleans the C* archive cache directory:

https://github.com/hawkular/hawkular/pull/812/files#diff-abe71b5f0a9948943e50733f447f83e9R35

ppalaga:
f4216c3 is good to merge from my PoV. The first Travis execution failed on the `HystrixCommandsITest` which is an indpendent problem, IMO. I restarted Travis.

@tsegismont Your proposal for removing just `"$HOME/.m2"` was accepted. Could you please review and eventually merge?

tsegismont:
@ppalaga LGTM. I agree, `HystrixCommandsITest` is certainly a different issue.

PR #813, opened 2016-01-28
HAWKULAR-984 : Update hawtio-oauth, skip KCAuthInterceptor for templates

ammendonca:
Also get rid of hawtio-template-cache as it doesn't seem required anymore.
Update charts version to fix some issues.


mtho11:
Merged #813

PR #820, opened 2016-02-01
HAWKULAR-784 - Enable compression for serving the Hawkular UI from Undertow

jmazzitelli:
…dertow

### DIFF
--- a/dist/src/main/resources/standalone.xsl
+++ b/dist/src/main/resources/standalone.xsl
@@ -220,7 +220,7 @@
   <xsl:template match="undertow:subsystem/undertow:server/undertow:host">
     <xsl:copy>
       <xsl:apply-templates select="node()|@*"/>
-      <filter-ref name="gzipFilter" predicate="path-suffix['.css'] or path-suffix['.js']"/>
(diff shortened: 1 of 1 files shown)


dist/src/main/resources/standalone.xsl
```diff
@@ -216,6 +216,21 @@
     <xsl:call-template name="loggers"/>
   </xsl:template>
 
+  <!-- Enable gzip compression when serving javascript and ccs -->
+  <xsl:template match="undertow:subsystem/undertow:server/undertow:host">
+    <xsl:copy>
+      <xsl:apply-templates select="node()|@*"/>
+      <filter-ref name="gzipFilter" predicate="path-suffix['.css'] or path-suffix['.js']"/>
```
mtho11: What about *.json and *.html?

jmazzitelli: We can add those additional path-suffix values, and anything else we deem necessary.

mtho11:
LGTM: Looks like it took the main hawkular js codebase: hawkular-console.js from 444 Kb to 72 Kb. 
Not too shabby. :+1:

mtho11:
Merged #820

PR #822, opened 2016-02-01
HAWKULAR-996 : Remove hk-alert-notification from URL details page

ammendonca:
Also fix the path to the directive template.
It should evaluated if this directive, inline notification, is to be used. If not, remove it completely from alerts.


mtho11:
Merged #822

PR #823, opened 2016-02-02
Migrate the tests to TestNG and Java

ppalaga:


ppalaga:
@Jiri-Kremser could you please review and eventually merge?

modules/end-to-end-test/src/test/java/org/hawkular/integration/test/HystrixCommandsITest.java
```diff
@@ -0,0 +1,154 @@
...
+//            Assert.fail("fail: " + e.getMessage())
+//        })
+//        println "observable pipes initialized";
+//
+//        // wait for the pipe
+//        def String urlJsonList = bigObs.toBlocking().first();
+//
+//        // check that nothing is there
+//        assertEquals("[ ]", urlJsonList);
+    }
```
Jiri-Kremser: You just commented out the hystrix test. That's not really the rewrite to TestNG :] Honestly, I don't see a value in rewriting it to TestNG from Groovy, but if you want to do that, I don't care. I am pretty indifferent of the testing framework, but I don't want to spend time on it, because it doesn't bring any value (imho). Rewriting it from 1 language to another 1..

It should be pretty straightforward, though.

`{foo -> bar}` Groovy lambdas should be replaced with `(foo -> bar)` Java lambdas,
perhaps the logic for getting the id from the location header may be slightly different,
otherwise all the method should be the same

ppalaga: Ah, sorry, I overlooked my own temporary comment-out. Thanks for catching it!

ppalaga:
@Jiri-Kremser 2f4f9de activates the missing part of the test.

Jiri-Kremser:
:+1: lgtm

PR #828, opened 2016-02-03
HAWKULAR-1009 : Adapt to metrics API change to percentiles

ammendonca:
Metrics have changed the way percentiles are requested and the way they
are sent. This was causing URL Percentil 95th to be shown as 0.

Also change chart type for RT from deprecated "hawkularmetric" to "line"


mtho11:
Merged #828

PR #829, opened 2016-02-03
Equal height in tokens cards

cardosogabriel:
Using flexbox. Looks good in Safari, Chrome and Firefox for Mac :)


mtho11:
Our first application of flexbox, Nice!

PR #832, opened 2016-02-04
HAWKULAR-1019 : Fix sparkline chart when c3-danger class is applied

ammendonca:
The fill should only be applied to sparklineArea and not to sparklineLine


mtho11:
Merged #832

PR #833, opened 2016-02-05
HAWKULAR-1020 - Cleaned and optimized some dependencies for modules.

jpkrohling:
I think I'd need the following people to provide input (testing to see if things are not broken, seeing if anything else can be removed, ...):

@ppalaga , @Jiri-Kremser , @pilhuhn 

If there's someone else who needs to review/test this, feel free to mention :)

### DIFF
--- a/modules/avail-creator/pom.xml
+++ b/modules/avail-creator/pom.xml
@@ -30,11 +30,18 @@
   <artifactId>hawkular-avail-creator</artifactId>
   <packaging>war</packaging>
 
-  <description>Component that takes status codes and creates avail messages
+  <name>Hawkular: Availability Creator</name>
+  <description>
+    A component that transforms HTTP status codes to UP and DOWN availability constants. Input are messages comming
+    via HawkularMetricData JMS topic and the output is sent to Metrics via REST.
   </description>
 
   <dependencies>
-
+    <dependency>
+      <groupId>org.hawkular.commons</groupId>
+      <artifactId>hawkular-bus-common</artifactId>
+      <scope>provided</scope> <!-- the nest provides this -->
+    </dependency>
 
     <dependency>
       <groupId>org.jboss.spec.javax.ejb</groupId>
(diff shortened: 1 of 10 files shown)


jpkrohling:
The "description"-related comments were ack'ed and fixed. I'll do the re-ordering, if that's consensus.

ppalaga:
> re-ordering, if that's consensus

It is not, although I'd be happy if it became one :)

ppalaga:
First review round finished.

jpkrohling:
Let me rephrase then: as I'm not the maintainer of those modules, I don't feel comfortable in taking a position here. So, I'll re-order and keep whatever ordering the maintainer wants :)

jpkrohling:
I believe that all items from the review were done, except for the reordering of the dependencies.

Jiri-Kremser:
Looks good in general. 
I've got couple of things:
1) thanks for doing it :+1: 
2) I know we should be using the jboss logging stuff, but the slf4j-simple is there for hystrix to do it's internal logging, if it wasn't there there was a message in the server log, that the Hystrix logging hasn't been set up
3) last time I tried to call the api, it wasn't working with the basic auth. 

this
```bash
curl -X POST -ivX -u jdoe:password -H "Content-Type: application/json" -d '{url: "www.seznam.cz"}' 'http://localhost:8080/hawkular/api/urls/' | less
```
should be passing, but wasn't because of 302 to login page (iirc)

I'll try to run the hawkular on this PR and smoke test.

jpkrohling:
>  2) I know we should be using the jboss logging stuff, but the slf4j-simple is there for hystrix to do it's internal logging, if it wasn't there there was a message in the server log, that the Hystrix logging hasn't been set up

I believe that you should need only the `slf4j-api`. Isn't the `-simple` an implementation of that API? If so, then Wildfly provides one implementation already (`logback`?) that directs the output to the `server.log`. But if you indeed have troubles with it, we can dig deeper.

jpkrohling:
@ppalaga , @Jiri-Kremser , @pilhuhn  : I changed a couple of things based on the reviews. Could you please re-check and re-test? Once you give me your approvals, I'll remove the "Don't merge yet" label.

Jiri-Kremser:
..looks ok to me

ppalaga:
LGTM

PR #834, opened 2016-02-05
Updated Keycloak JavaScript Adapter to 1.8.1

jpkrohling:


jpkrohling:
The build currently has problems with the Angular version. Once it's fixed, the Travis for this has to be re-executed. This version is the same as the one we have in the backend and should get rid of the "deprecated endpoint" warnings.

mtho11:
Merged #834

PR #836, opened 2016-02-08
more informative itests failures 

Jiri-Kremser:
_400 should be 200' is not very helpful_ - (https://travis-ci.org/hawkular/hawkular/builds/107327309#L2328)


Jiri-Kremser:
ok, now we are smarter:

java.lang.AssertionError: Response msg: {"errorMsg":"Tenant is not specified. Use 'Hawkular-Tenant' header."} expected:<200> but was:<400>

pilhuhn:
I would have liked you writing "no more iTest failures" :-)

PR #839, opened 2016-02-09
HAWKULAR-960 Create message center.

karelhala:
Add all toastr messages to this center. Create loading message, which is responsible for showing that something is happening on server. Add action message to message center, this message shows custom html which is for taking actions, either download button or redirect etc. Add show more button, add number of messages which were not yet read

HAWKULAR-960 fix registering of watchers and listeners.
Watcher for experimental mode was added all the time to , this might lead to some memory problems. Also JDR export success and error were reigstered in same way.

### DIFF
--- a/console/src/main/scripts/plugins/metrics/ts/appServerDetails.ts
+++ b/console/src/main/scripts/plugins/metrics/ts/appServerDetails.ts
@@ -41,27 +41,27 @@ module HawkularMetrics {
     /* tslint:disable:variable-name */
 
     constructor(private $rootScope: any,
-      private $scope: any,
-      private $route: any,
-      private $routeParams: any,
-      private $q: ng.IQService,
-      private $timeout: ng.ITimeoutService,
-      private HawkularOps: any,
-      private NotificationsService: INotificationsService,
-      private HawkularInventory: any,
-      private HawkularAlertsManager: any,
-      private $log: ng.ILogService,
-      private $location: ng.ILocationService,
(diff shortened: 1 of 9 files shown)


console/src/main/scripts/plugins/metrics/ts/appServerDetails.ts
```diff
@@ -685,14 +706,14 @@ module HawkularMetrics {
           .withName('CPU usage')
           .withDescription(description)
           .withActions(this.defaultAction)
-          .withTags({ resourceId: this.feedId, resourceName: cpuResource['name'] })
+          .withTags({resourceId: this.feedId, resourceName: cpuResource['name']})
           .withContext(
-          new AlertDefinitionContextBuilder(AlertDefinitionContext.TRESHOLD_TRIGGER_TYPE)
```
mtho11: Please make the builders static. We shouldn't have to **new** an instance every time, and it just creates extra instances for no benefit (introduces unnecessary GC).

karelhala: How about we create new task for refactoring these builders. Since it is not exactly the point of this issue and it might take some time to refactor all those builders.

mtho11: Sounds fine. Here ya go: https://issues.jboss.org/browse/HAWKULAR-1030

mtho11:
Merged #839

PR #840, opened 2016-02-09
Accounts 2.0.21.Final

jpkrohling:


jpkrohling:
This required ui-services 0.9.1, which is not released yet.

@ppalaga , @metlos , could you please review the feature-pack part? I'm not very comfortable in adding the Keycloak dependency here, but it seems to be required. 

@mtho11 , could you review the UI parts?

ppalaga:
> I'm not very comfortable in adding the Keycloak dependency here, but it seems to be required.

There is a good hope that the KC version prop will be only temporarily here.

The PR looks good to me.

jpkrohling:
@karelhala , @ammendonca , @vrockai are also welcome to review the UI parts :)

PR #842, opened 2016-02-10
HAWKULAR-1027 change creating of action in trigger based on new API

karelhala:
New API for actions is array of objects, so correspond with them. Create new builder for this. For now don't store properties in since no properties are available in constructor of new action in trigger


mtho11:
Merged #842

PR #845, opened 2016-02-11
HAWKULAR-1035 : Fix mismatch in DS used percentage

ammendonca:


mtho11:
Merging #845 despite the iTest failures that have nothing to do with UI code.

PR #846, opened 2016-02-12
Hawkular 922

jshaughn:
Migrate trigger definitions from UI to server listener.  And move to group triggers in order to manage out-of-box alerts at a template level.


jshaughn:
Capturing comments from bluejeans review with Lucas:
- remove some commented/dead code in alertsManager.ts
- remove some console logging as well
- test that new e-mail scheme is working end-to-end
- future: consider using XML file for the group trigger definitions in the listener

jshaughn:
From Lucas review:
Perhaps the member trigger names should not be the same as the group trigger name, but add some level of distinction (but shorter than the more verbose description)

mtho11:
This might be better as a squash merge.

jshaughn:
@mtho11 When ready with some PR feedback changes, later today, if you walk me through the squash I will do it. I won't try and do a squash on my own.

modules/hawkular-api-parent/hawkular-listener/src/main/java/org/hawkular/listener/bus/InventoryEventListener.java
```diff
@@ -0,0 +1,889 @@
...
+        try {
+            init();
+
+            String tenantId = event.getTenant().getId();
+            ResourceType rt = event.getObject();
+            String type = rt.getId();
+            switch (type) {
+                case "Datasource": {
+                    // Available Connections
+                    String groupTriggerId = "DS_Conn";
```
mtho11: What about moving these lines of code to a separate method, there are lots of lines of code in this case block.

console/src/main/scripts/plugins/metrics/ts/triggers/eventTrigger.ts
```diff
@@ -33,15 +33,25 @@ module HawkularMetrics {
...
+          if ( this.adm.trigger['email'] === undefined || this.adm.trigger['email'] === null ) {
+            this.adm.trigger['emailEnabled'] = false;
+            this.adm.trigger['email'] = this.$rootScope.userDetails.email;
+          } else {
+            this.adm.trigger['emailEnabled'] = true;
+          }
 
           // presentation
           this.adm.trigger['context'] = triggerData.trigger.context;
+          this.adm.trigger['description'] = triggerData.trigger.description;
```
mtho11: In the future, try not to use **trigger.['description']** because it bypasses type checking, but instead use **trigger.description** (which the IDE will also autocomplete for you).

modules/hawkular-api-parent/hawkular-listener/src/main/java/org/hawkular/listener/bus/InventoryEventListener.java
jshaughn: I knew someone wouldn't like my huge switch stmt :)  I can break it up but will do so after the PR goes in.

console/src/main/scripts/plugins/metrics/ts/triggers/eventTrigger.ts
jshaughn: Alex already mentioned this style point as well.  This was existing code that was based on other existing code I had seen.  I can do the conversion, but after this PR goes in.

mtho11:
Merged #846

PR #847, opened 2016-02-12
HAWKULAR-1032 - Pinger now sends data to metrics via bus

jpkrohling:


jpkrohling:
This requires hawkular/hawkular-metrics#445 . Few things to note:

- The mentioned PR for metrics has to be locally built, and then build this (as it contains a reference to the snapshot). Once that PR is merged, we can use src-deps.
- The `MetricsDataMessage` exists in Metrics and Alerts already. None of which we want a direct dependency on pinger. A proper solution would be a more generic place to put payloads, but until then, I think it's OK to live with this duplication of code.
- The queue names also come from the Metrics PR. I'm assuming those will not change for now.
- I tested locally that data is sent to metrics and that metrics receives it. I could *not* check the UI, though. Not sure if there are changes required on that part. 

So, I'd need a review from:

@tsegismont for the metrics-related parts
@ppalaga or @pilhuhn , the owners of pinger
@ammendonca , @karelhala or @mtho11 for the UI part, to assess if changes are needed there.

jpkrohling:
PR updated to make metrics point to the src-deps of the current head of master, as the dependent feature has been merged.

ppalaga:
LGTM

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricDataMessage.java
```diff
@@ -24,10 +25,16 @@
...
 
 /**
+ * A bus message for messages on HawkularMetricData Topic.
+ *
+ * @author Jay Shaughnessy
  * @author Lucas Ponce
  */
+
 public class MetricDataMessage extends AbstractMessage {
+    // TODO: we now have *3 copies* of this payload... shouldn't we have a place for payloads?
```
tsegismont: +1

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
```diff
@@ -42,21 +35,11 @@
...
-        outer.put("data", data);
-        mMetrics.add(outer);
-    }
-
-    private final PingerConfiguration configuration = PingerConfiguration.getInstance();
+    @Resource(mappedName = "java:/HawkularBusConnectionFactory")
+    private ConnectionFactory connectionFactory;
 
     /**
      * Serializes the given {@link PingStatus} and then submits it to Hawkular-metrics service via REST
```
tsegismont: Javadoc needs to be deleted or updated :)

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
```diff
@@ -65,39 +48,39 @@ private static void addDataItem(List<Map<String, Object>> mMetrics, String resou
...
+        message.setMetricData(metricData);
 
-        try {
-            HttpResponse response = client.execute(request);
-            if (response.getStatusLine().getStatusCode() > 399) {
-                Log.LOG.wMetricPostStatus(response.getStatusLine().toString());
-            }
-        } catch (IOException e) {
-            Log.LOG.eMetricsIoException(e);
+        try (JMSContext context = connectionFactory.createContext()) {
```
tsegismont: Minor: I believe we could avoid to create a context and producer any time a ping status is sent

tsegismont:
LGTM

I don't believe the UI part is impacted by this PR. You only changed the way the Pinger reports data to Metrics.

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
jpkrohling: Oops, good catch!

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
jpkrohling: Not sure about that: ActiveMQ closes the context automatically if we don't, so, we might end up with a stale context. Using this try-with-resources, we ensure that we are closing, but that's pretty much all we can do. I'd love to be proven wrong, though.

jpkrohling:
Right, the UI part is to make sure I didn't break anything. I can't check that, as the UI is "broken" already.

tsegismont:
Le 15/02/2016 16:10, Juraci Paixão Kröhling a écrit :
> Right, the UI part is to make sure I didn't break anything. I can't
> check that, as the UI is "broken" already.

I saw that @jkremser sent a PR to hawkular-ui-services already. The 
remaining part is related to hawkular-charts. I'll check with @mithomps ASAP

modules/pinger/src/main/java/org/hawkular/component/pinger/MetricPublisher.java
tsegismont: Le 15/02/2016 16:09, Juraci Paixão Kröhling a écrit :
> Not sure about that: ActiveMQ closes the context automatically if we
> don't, so, we might end up with a stale context. Using this
> try-with-resources, we ensure that we are closing, but that's pretty
> much all we can do. I'd love to be proven wrong, though.

https://github.com/hawkular/hawkular-metrics/blob/master/hawkular-component/src/main/java/org/hawkular/metrics/component/insert/NewDataListener.java#L86

tsegismont: See also http://activemq.apache.org/how-do-i-use-jms-efficiently.html

Written for ActiveMQ but still relevant for Artemis

tsegismont: But this is not blocking AFAIK

PR #849, opened 2016-02-15
HAWKULAR-1026 unshallow the git clone so that license plugin stopps flagging fil

ppalaga:


ppalaga:
@pilhuhn could you please review and eventually merge? Looks like I was blacklisted by Travis because have sent too many PRs within a short period of time.

PR #854, opened 2016-02-16
HAWKULAR-1040 Issue with URL trigger definition

jshaughn:
- handle the fact that the actual URL is not supplied in
  the resource create event published by inventory.

### DIFF
--- a/modules/hawkular-api-parent/hawkular-listener/src/main/java/org/hawkular/listener/bus/InventoryEventListener.java
+++ b/modules/hawkular-api-parent/hawkular-listener/src/main/java/org/hawkular/listener/bus/InventoryEventListener.java
@@ -696,14 +696,10 @@ public class InventoryEventListener extends InventoryEventMessageListener {
                 }
                 case "URL": {
                     // common to members
-                    String feedId = r.getPath().ids().getFeedId();
-                    String url = event.getHeaders().get("url");
-                    //String resourceId = String.valueOf(MessageDigest.getInstance("MD5").digest(
-                    //        r.getId().getBytes("UTF-8")));
+                    // TODO: use the URL as needed, not the hash if and when we can get the actual URL from the event
                     String resourceId = r.getId();
-                    //String qualifiedResourceId = feedId + "/" + resourceId;
                     Map<String, String> memberContext = new HashMap<>(2);
-                    memberContext.put("resourceName", url);
+                    memberContext.put("resourceName", "URL [" + resourceId + "]"); // Should be the URL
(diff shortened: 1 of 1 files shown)


modules/hawkular-api-parent/hawkular-listener/src/main/java/org/hawkular/listener/bus/InventoryEventListener.java
```diff
@@ -696,14 +696,10 @@ private void handleResourceEvent(ResourceEvent event) {
...
-                    String feedId = r.getPath().ids().getFeedId();
-                    String url = event.getHeaders().get("url");
-                    //String resourceId = String.valueOf(MessageDigest.getInstance("MD5").digest(
-                    //        r.getId().getBytes("UTF-8")));
+                    // TODO: use the URL as needed, not the hash if and when we can get the actual URL from the event
                     String resourceId = r.getId();
-                    //String qualifiedResourceId = feedId + "/" + resourceId;
                     Map<String, String> memberContext = new HashMap<>(2);
-                    memberContext.put("resourceName", url);
+                    memberContext.put("resourceName", "URL [" + resourceId + "]"); // Should be the URL
```
lucasponce: url == resourceId is confirmed in all cases ?

modules/hawkular-api-parent/hawkular-listener/src/main/java/org/hawkular/listener/bus/InventoryEventListener.java
```diff
@@ -711,8 +707,7 @@ private void handleResourceEvent(ResourceEvent event) {
                     // Response Time
                     String groupTriggerId = "URL_Response";
                     String memberId = groupTriggerId + "_" + resourceId;
-                    String memberName = "URL Response [" + url + "]";
```
lucasponce: So, perhaps a memberName should be good, to not leave this trigger without name.

jshaughn: It defaults to the group trigger name, which is currently the policy for all of the member triggers.

modules/hawkular-api-parent/hawkular-listener/src/main/java/org/hawkular/listener/bus/InventoryEventListener.java
jshaughn: The actual URL is not supplied by inventory, so using the resourceId is a workaround until such time that we have the URL.

lucasponce:
LGTM

PR #865, opened 2016-02-29
HAWKULAR-1053 change loading of recursive children and metrics

karelhala:
When recursive children are loaded slash needs to be hashed to %2F. When metrics are loaded for selected resource whole path from feed id should be taken, not only id.


mtho11:
Merged #865

PR #866, opened 2016-03-01
HAWKULAR-1049 change to date on overview page to show today/yesterday

karelhala:
if selected range of time is in yesterday/today, show this in overview page, if the date is older than yesterday show actual date. In inventory explorer automatically update time under timepicker when refreshed (either manualy or automatically).

### DIFF
--- a/console/src/main/scripts/plugins/metrics/ts/directives/overview/statusOverviewDirective.ts
+++ b/console/src/main/scripts/plugins/metrics/ts/directives/overview/statusOverviewDirective.ts
@@ -20,6 +20,28 @@
 
 module HawkularMetrics {
   export class StatusOverviewController {
+    constructor(private $filter: ng.IFilterService) {}
+
+    public parseDayFromTimestamp(timestamp): string {
+      const dateFilter = this.$filter('date');
+      const parsedDate = new Date(timestamp);
+      const today = new Date();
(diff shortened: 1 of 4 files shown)


console/src/main/scripts/plugins/metrics/ts/directives/overview/statusOverviewDirective.ts
```diff
@@ -20,6 +20,28 @@
...
   export class StatusOverviewController {
+    constructor(private $filter: ng.IFilterService) {}
+
+    public parseDayFromTimestamp(timestamp): string {
+      const dateFilter = this.$filter('date');
+      const parsedDate = new Date(timestamp);
+      const today = new Date();
+      const yesterday = new Date();
+      yesterday.setDate(today.getDate() - 1);
+      if (StatusOverviewController.isSameDayMonthYear(parsedDate, yesterday)) {
```
jpkrohling: Wouldn't it be better to use Moment.js for that? I think there's already a dependency on it.

karelhala: Yeah I guess this might be better in moment.js
I will change it and reopen PR once egain.

mtho11:
Merged #866

PR #867, opened 2016-03-01
HAWKULAR-1049 change from plain Date object to moment

karelhala:


console/src/main/scripts/plugins/metrics/ts/directives/overview/statusOverviewDirective.ts
```diff
@@ -23,26 +23,17 @@ module HawkularMetrics {
...
-      yesterday.setDate(today.getDate() - 1);
-      if (StatusOverviewController.isSameDayMonthYear(parsedDate, yesterday)) {
-        return 'Yesterday';
-      } else if (StatusOverviewController.isSameDayMonthYear(parsedDate, today)) {
+      const parsedDate = moment(timestamp).startOf('day');
+      const today = moment().startOf('day');
+      if (today.diff(parsedDate) === 0) {
         return 'Today';
+      } else if (today.add(-1, 'days').startOf('day').diff(parsedDate) === 0) {
+        return 'Yesterday';
```
jpkrohling: Moment.js has a `calendar()` method, which should render "Today", "Yesterday" and "Tomorrow" as needed: http://momentjs.com/docs/#/displaying/calendar-time/

karelhala: Well the `calendar()` is used like: `moment().add(-1, 'days').calendar()` and the output is `Yesterday at 4:28 PM` I don't think we want PM/AM in there. But I can still change settings of calendar and use this instead returning string 'Yesterday' or 'Today'

jpkrohling: Alright, I thought it would take the user locale into consideration and use 24h-format accordingly. If that's not the case, then makes sense to do it the way you did :)

mtho11:
Merged #867

PR #868, opened 2016-03-03
HAWKULAR-1037 update UI to be more aligned

karelhala:
Move resource selector and feed selector to the center of page. Move resource tree to it's own component

Remove funcion, we use ES6 coding

### DIFF
--- a/console/src/main/scripts/plugins/metrics/html/directives/explorer/resource-tree.html
+++ b/console/src/main/scripts/plugins/metrics/html/directives/explorer/resource-tree.html
@@ -1,0 +1,6 @@
+<div class="hk-resource-tree">
+  <label>Resource children tree</label>
+  <div ng-if="$ctrl.childrenTree.length !== 0">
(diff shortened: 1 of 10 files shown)


mtho11:
Merged #868

console/src/main/scripts/plugins/metrics/html/directives/explorer/resource-tree.html
```diff
@@ -0,0 +1,6 @@
+<div class="hk-resource-tree">
+  <label>Resource children tree</label>
+  <div ng-if="$ctrl.childrenTree.length !== 0">
+    <hk-boot-nav-tree tree-data="$ctrl.childrenTree" on-node-selected="$ctrl.onLoadMetrics({resource: resource})"></hk-boot-nav-tree>
```
mtho11: The controllerAs label `$ctrl` is not very descriptive (they are all controllers) and the `$` makes it feel like its a special Angular variable that shouldn't be touched.

PR #871, opened 2016-03-09
HAWKULAR-1054 collapse option for graph selector

karelhala:
Resource, feed and metric can be now collapsed with clicking on header

Make graph selector hidden by default


pilhuhn:
LGTM :+1:

mtho11:
Merged #871

PR #873, opened 2016-03-15
If no tag matches, just emit some default message.

pilhuhn:
@jshaughn @mtho11


mtho11:
Merged #873 despite Travis Integration tests failing (like they do much of the time).

PR #877, opened 2016-03-22
HAWKULAR-1059 : Fix deployments select on refresh

ammendonca:
When a periodic refresh occurred, the deployments table was selecting
all the items in the table. This fixes that behavior.


mtho11:
Merge #877

PR #879, opened 2016-04-06
Shrinkwrap the npm package.json locking the versions from unforeseen …

mtho11:
…changes (and breaking the build). Because of the transitive dependencies, even a fixed versioning mechanism is not enough to protect against shifting versions and breakage.Npm-shrinkwrap locks all of the dependent versions for repeatable builds.


mtho11:
Merged #879

PR #882, opened 2016-05-19
Charts removed height property

pavolloffay:
@mtho11 could you pleser review? Yesterday we missed to remove some height properties.


mtho11:
LGTM, the chart height property is not used now since its responsive, it uses the chart height of its container. It will fill whatever the container size is now.

